PR #11, opened 2020-12-10
Simple Berlin-Brandenburg model

### LINKED ISSUE #2: There is no basic model
Transfer basic oemof app from flexmex.

c-moeller:
This adds a simple model for the Berlin-Brandenburg region as a base for further development.

Fixes #2

TODO:
* [x] Change scenario names to make them more descriptive Any suggestions, @c-moeller?
* [x] Enable processing in the snakemake pipeline of oemof-b3
* [x] Check postprocessed data (possible after merging #23) 
* [x] Finalize documentation #27


c-moeller:
Ok, I fixed this bug. I forgot to fill some data.

Now I get another error when I run optimize.py. Could this be an issue with the time index in my profile file?

```
Traceback (most recent call last):
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/oemof/tabular/datapackage/reading.py", line 294, in deserialize_energy_system
    facade_data = r.read(keyed=True, relations=True)
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/datapackage/resource.py", line 385, in read
    foreign_keys_values=foreign_keys_values, **options)
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/tableschema/table.py", line 353, in read
    for count, row in enumerate(rows, start=1):
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/tableschema/table.py", line 282, in iter
    error_data=local_keyed_values)
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/tableschema/helpers.py", line 90, in default_exc_handler
    raise exc
tableschema.exceptions.UnresolvedFKError: Foreign key "['profile']" violation in row "2": ('BE-wind-onshore-profile',) not found in wind-onshore_profile

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "optimize.py", line 25, in <module>
    attributemap={}, typemap=TYPEMAP,
  File "/home/caroline/.virtualenvs/oemof-B3/lib/python3.6/site-packages/oemof/tabular/datapackage/reading.py", line 301, in deserialize_energy_system
    "Exception was: {}").format(r.name, e))
datapackage.exceptions.LoadError: Could not read data for resource with name `wind-onshore`.  Maybe wrong foreign keys?
Exception was: Foreign key "['profile']" violation in row "2": ('BE-wind-onshore-profile',) not found in wind-onshore_profile
```

examples/simple_model/preprocessed/simple_model/data/elements/electricity-curtailment.csv
```diff
@@ -0,0 +1,3 @@
+region,name,type,carrier,tech,bus,capacity,marginal_cost,output_parameters
+BE,BE-electricity-curtailment,excess,electricity,curtailment,BE-electricity,0,0,{}
```
jnnr, line 2: Capacity should not be zero, should it?

examples/simple_model/preprocessed/simple_model/data/elements/electricity-shortage.csv
```diff
@@ -0,0 +1,3 @@
+region,name,type,carrier,tech,bus,capacity,marginal_cost,output_parameters
+BE,BE-electricity-shortage,shortage,electricity,shortage,BE-electricity,0,0,{}
```
jnnr, line 2: Capacity = 0 ?

oemofB3/model_structure/busses.csv
```diff
@@ -1,5 +1,6 @@
 carrier,balanced,description
 electricity,true,Electricity bus
-heat,true,Heat bus
-ch4,false,CH4 bus
-uranium,false,Uranium bus
\ No newline at end of file
+heat,false,Heat bus
```
jnnr, line 3: Would be more consistent to make all buses balanced and include a heat-shortage.

oemofB3/model_structure/component_attrs/biomass-st.csv
```diff
@@ -0,0 +1,15 @@
...
+region,str,n/a,n/a,,Country code
+name,str,n/a,,-biomass-st,Name
+type,str,n/a,conversion,,Type
+carrier,str,n/a,biomass,,Energy carrier
+tech,str,n/a,st,,Technology
+from_bus,str,n/a,,-biomass,The input bus this component is connected to
+to_bus,str,n/a,,-electricity,The output bus this component is connected to
+capacity,float,MW,n/a,,Installed capacity
+capacity_cost,float,Eur/MW,n/a,,Annualized costs of capacity expansion
+efficiency,float,unitless,n/a,,Efficiency of the conversion ch4 to electricity
```
jnnr, line 11:
```suggestion
efficiency,float,unitless,n/a,,Efficiency of the conversion biomass to electricity
```

oemofB3/model_structure/component_attrs/oil-st.csv
```diff
@@ -0,0 +1,15 @@
...
+region,str,n/a,n/a,,Country code
+name,str,n/a,,-oil-st,Name
+type,str,n/a,conversion,,Type
+carrier,str,n/a,oil,,Energy carrier
+tech,str,n/a,st,,Technology
+from_bus,str,n/a,,-oil,The input bus this component is connected to
+to_bus,str,n/a,,-electricity,The output bus this component is connected to
+capacity,float,MW,n/a,,Installed capacity
+capacity_cost,float,Eur/MW,n/a,,Annualized costs of capacity expansion
+efficiency,float,unitless,n/a,,Efficiency of the conversion ch4 to electricity
```
jnnr, line 11:
```suggestion
efficiency,float,unitless,n/a,,Efficiency of the conversion oil to electricity
```

oemofB3/model_structure/component_attrs/other-st.csv
```diff
@@ -0,0 +1,15 @@
...
+region,str,n/a,n/a,,Country code
+name,str,n/a,,-other-st,Name
+type,str,n/a,conversion,,Type
+carrier,str,n/a,other,,Energy carrier
+tech,str,n/a,st,,Technology
+from_bus,str,n/a,,-other,The input bus this component is connected to
+to_bus,str,n/a,,-electricity,The output bus this component is connected to
+capacity,float,MW,n/a,,Installed capacity
+capacity_cost,float,Eur/MW,n/a,,Annualized costs of capacity expansion
+efficiency,float,unitless,n/a,,Efficiency of the conversion ch4 to electricity
```
jnnr, line 11:
```suggestion
efficiency,float,unitless,n/a,,Efficiency of the conversion of 'other' to electricity
```

examples/simple_model/preprocessed/simple_model/data/elements/electricity-curtailment.csv
c-moeller: I couldn't figure out what's meant with capacity in case of curtailment. So I thought, I set it (the initial capacity?) to zero and the optimizer can expand it.

examples/simple_model/preprocessed/simple_model/data/elements/electricity-shortage.csv
c-moeller: Same here... but maybe I got it wrong.

oemofB3/model_structure/busses.csv
c-moeller: Ah, thats a mistake. I remove the heat bus because I don't have one. I thougth, the 'False' stands for existing or not.

examples/simple_model/preprocessed/simple_model/data/elements/electricity-curtailment.csv
jnnr: You could just not set capacity, so curtailment is unrestricted.

oemofB3/model_structure/busses.csv
jnnr: I see! You could remove it, but this would affect all scenarios.

jnnr:
Looks good and runs, once the right package versions are installed. The scripts "setup_model_structure" could be removed and maybe also "infer", as the datapackage is already provided.

TODO
* [ ] Short documentation: What are these three scenarios?
* [ ] Specify oemof-tabular dev branch as requirement (released version on pip throws foreign key error)

oemofB3/model_structure/busses.csv
jnnr: I made the heat bus balanced again in 70c56c3 and did not remove it, as it will be part of the model later. The resource busses have to be unbalanced because there are no commodity sources defined. There is no way to exclude unused buses at the moment depending on the scenario. I opened an issue (#18)

jnnr:
I corrected the bus definition in all 3 example models to make the commodity busses unbalanced. This should affect the results, so keep in mind to re-run the model before interpreting results.

jnnr:
We had a look at the results plots and found a bug: There seems to be a bug with the time series data for renewables. PV, e.g., should be normalized, but features some values larger than 600, e.g. at 2019-04-23 13:00:00. Can you help us here, @c-moeller?

c-moeller:
It seems like an error with the decimal point which might have happend due to formatting. I can try to find the original time series. Another option would be to integrate the feedinlib. What do you think?

jnnr:
@c-moeller: Another question concerning the scenarios: Currently, biomass and li-ion batteries are expandable, all others not. The docs tell another story.

From my view, it would make more sense to keep all capacities fixed in the scenarios where you take the capacities from NEP. We could introduce a 4th scenario that does a greenfield optimisation to show how that works.

jnnr:
I updated the scripts for setting up the model structure and optimization

* Only one script for all three to avoid code duplication
* Use the EnergyDataPackage class of oemoflex for setting up defaults

TODO

* [ ] Allow to include oil-st, other-st in oemoflex by allowing to pass extra components.

c-moeller:
> @c-moeller: Another question concerning the scenarios: Currently, biomass and li-ion batteries are expandable, all others not. The docs tell another story.
> 
> From my view, it would make more sense to keep all capacities fixed in the scenarios where you take the capacities from NEP. We could introduce a 4th scenario that does a greenfield optimisation to show how that works.

Ok, sounds reasonable. But I think, I would at least leave the battery capacity open, because there are no data in the NEP for batteries (as far as I remember), thus they are not preinstalled. The problem could be infeasible. But we can also try to leave the battery out. It could also work with grid transfer and dispatchable fossils.

Maybe we add two scenarios to the existing 3:
4) with additional batteries
5) greenfield

jnnr:
The last 5 commits fix two things: 
- Change the scenario names in the scripts
- Make the scenarios feasible by adapting capacity of shortage and curtailment
- set cost on shortage.

juliusmeier:
The revision of the examples scenarios included:
- renaming scenarios: 
simple_model -> base (A 2035) 
simple_model_2 -> more_renewables (B 2035)
simple_model_3 -> more_renewables_less_fossil (B 2040)
- adapt values in csv-tables to match NEP
- capacity_cost, storage_capacity_cost empty
- expendable = False

Open question:
- [ ]  What to do with the liion-batteries? 
NEP lists "Weitere Speicher und nachfrageseitige Flexibilitäten [GW]" for every scenario on page 23
- [ ] Examples scenario names might be renamed again because the NEP scenarios rather focus on increasing sector coupling/electrification and grid orientation as shown on page 22.

jnnr:
In the results, the transmission capacity constraint is not respected. Probably because the attributes of the `Link` class in oemof.tabular have changed in the meantime. 'capacity' has been replaced by 'from_to_capacity' and 'to_from_capacity' to enable different capacities depending on direction.

@juliusmeier: Can you adapt `electricity-transmission.csv` accordingly and try if that fixes the problem?

juliusmeier:
> In the results, the transmission capacity constraint is not respected. Probably because the attributes of the `Link` class in oemof.tabular have changed in the meantime. 'capacity' has been replaced by 'from_to_capacity' and 'to_from_capacity' to enable different capacities depending on direction.

Do I need to update my oemof.tabular package? How do I update the dev branch there?

juliusmeier:
> @juliusmeier: Can you adapt `electricity-transmission.csv` accordingly and try if that fixes the problem?

Yes, I fixed it with https://github.com/rl-institut/oemof-B3/pull/11/commits/6226a4c61ea9631e33dadc209b4cf572f6f0c67f

juliusmeier:
If you look on the plots, e.g. BE-electricity_01 in the base scenario, there are some timesteps where the demand is not served. Maybe you can have a look, @jnnr ?

jnnr (changes requested):
Just one thing, apart from that I approve.

jnnr (approved):

PR #22, opened 2021-02-15
VariationGenerator

jnnr:
This introduces a variation generator which creates new datapackages from existing ones by applying variations defined in a data table.


jnnr:
Closed, as the results of these developments have been moved to oemoflex (https://github.com/rl-institut/oemoflex/)

PR #23, opened 2021-02-15
Features/import oemoflex

jnnr:
This PR deletes the files responsible for model structure and imports all functionality from oemoflex (https://github.com/rl-institut/oemoflex). This makes sure that generic developments are separated from model specifics.


jnnr:
Tested successfully with example and basic scenario.

PR #20, opened 2021-02-05
Update Readme

monika-o:
Add some backgound information about the tool

PR #27, opened 2021-03-03
Monika's Patch 2

monika-o:
The aim of this branch is to add an example to the docs. I hope that it makes sense the ways I did it. I'm looking forward to your comments.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -1,4 +1,10 @@
-How It Works
+.. _examples_label:
+
+~~~~~~~~~~~~~~~
+Examples
+~~~~~~~~~~~~~~~
+
+Example 1
 =====
 The model consists of two nodes, one for Brandenburg and one for Berlin.
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -1,0 +1,33 @@
+Overview
+========
+The Open Energy Modelling Framework B3 (oemof-B3) is a simple model to map the electricity network in the area of Berlin-Brandenburg (Germany).
+It serves as a basis for further development.
+
+How It Works
+=====
+The model consists of two nodes, one for Brandenburg and one for Berlin.
+
+Modelled components are 
+
+- renewable energies (wind, photovoltaic, biomass)
+- conventional power plants (gas, oil, others)
+
+Data for capacities of wind, photovoltaic and biomass as well as electricity demand have been taken from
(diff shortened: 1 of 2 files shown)


jnnr:
@monika-o : Thanks for opening this. Is this branch supposed to be a replacement for #20? If so, we can document it here in the chat and close the other PR.

docs/examples.rst
```diff
@@ -0,0 +1,34 @@
...
+the grid development plan (`Netzentwicklungsplan <https://www.netzentwicklungsplan.de/sites/default/files/paragraphs-files/NEP_2035_V2021_1_Entwurf_Teil1.pdf>`_, p. 41 ff.).
+Data for efficiency and costs (specific annuity, fuel costs) are based on different sources as well as own assumptions and can be adjusted.
+
+
+All electricity generating components can be expanded. 
+Additionally, it can be invested into grid expansion and battery storage.
+
+Three scenarios have been built based on the grid development plan:
+
+- simple_model: corresponds to scenario A 2035 of the grid development plan
```
jnnr, line 26: We should think about names that are more telling.

jnnr:
We should mention the units of the input data (kW?)

docs/examples.rst
jnnr: @c-moeller: I wonder how close the correspondence to NEP is? Would it justify to call the scenarios by A-2035 etc.? Or would it be more honest to say that the scenarios have been setup with a loose correspondence to the scenarios in NEP by reflecting higher wind capacities and less ch4-gt in 2040? This should be consistent with the naming and explanations at OEP. There, the scenarios seem to be called scenario 1, scenario 2, scenario 3. Is that fixed, @christian-rli? In that case we should rename the scenarios here accordingly.

christian-rli: The naming is not fixed. In fact, we thought it would be useful to give more descriptive names, so feel free to rename wildly and let us know the result :)

jnnr: Good! Can you give some input here along the questions posed above or shall we discuss this, @c-moeller?

c-moeller: Does that help? Or do you need more information concerning the scenarios?

jnnr: We came up with this:
* basic
* more_re
* more_re_less_fossil

jnnr:
I merge this branch now as most of it is settled for now. Will process the remaining details on a separate one.

PR #29, opened 2021-04-15
Rule for scenario report

jnnr:
This is a draft for a snakemake rule that creates automatic reports for scenarios from plots.


juliusmeier:
I get this error. I already set the latency to 10s but the same error occured.
```
snakemake -j4 --keep-target-files report.pdf
```

```
The flag 'directory' used in rule prepare_example is only valid for outputs, not inputs.
The flag 'directory' used in rule optimize is only valid for outputs, not inputs.
Building DAG of jobs...
Provided cores: 4
Rules claiming more threads will be scaled down.
Job counts:
	count	jobs
	1	report
	1
Select jobs to execute...

[Wed May 26 16:34:04 2021]
rule report:
    input: report/report.md
    output: report.pdf
    jobid: 0

Waiting at most 5 seconds for missing files.
MissingOutputException in line 50 of C:\Users\meinm\Documents\Git\oemof-B3\Snakefile:
Job Missing files after 5 seconds:
report.pdf
This might be due to filesystem latency. If that is the case, consider to increase the wait time with --latency-wait.
Job id: 0 completed successfully, but some output files are missing. 0
  File "c:\users\meinm\anaconda3\envs\rli\lib\site-packages\snakemake\executors\__init__.py", line 584, in handle_job_success
  File "c:\users\meinm\anaconda3\envs\rli\lib\site-packages\snakemake\executors\__init__.py", line 252, in handle_job_success
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: C:\Users\meinm\Documents\Git\oemof-B3\.snakemake\log\2021-05-26T163404.232401.snakemake.log
```

jnnr:
> I get this error. I already set the latency to 10s but the same error occured.

This is because the output defined in Snakefile is not the same as the file that is produced by the pandoc command.

In the last commit, I adapted the rule a bit so it builds the report in the specific scenario directory. When #32 is merged, this allows the report rule to follow directly after plotting. 

There was a problem with including .pdf though, is that true?

juliusmeier:
> This is because the output defined in Snakefile is not the same as the file that is produced by the pandoc command. In that case, snakemake is waiting for the file to appear and complains as it does not. But the file is created by pandoc anyway.
> 
> In the last commit, I adapted the rule a bit so it builds the report in the specific scenario directory. When #32 is merged, this allows the report rule to follow directly after plotting.

Ah, I undertand. Unfortunately, there is a new problem now. Windows does not have ``cp`` as a shell command. It uses ``copy`` instead.


> There was a problem with including .pdf though, is that true?

Yes, but it is resolved. In markdown itself a workaround is needed to include .pdf images. But since pandoc uses some enhanced markdown, it is possible to include .pdf images just like other image files.

juliusmeier:
The report directory and file are created but snakemake still raises the error.
```
(rli) C:\Users\meinm\Documents\Git\oemof-B3>snakemake -j4 --keep-target-files results/simple_model/report
The flag 'directory' used in rule prepare_example is only valid for outputs, not inputs.
The flag 'directory' used in rule optimize is only valid for outputs, not inputs.
Building DAG of jobs...
Provided cores: 4
Rules claiming more threads will be scaled down.
Job counts:
        count   jobs
        1       report
        1
Select jobs to execute...

[Thu May 27 13:31:29 2021]
rule report:
    input: report/report.md
    output: results/simple_model/report
    jobid: 0

Waiting at most 5 seconds for missing files.
MissingOutputException in line 50 of C:\Users\meinm\Documents\Git\oemof-B3\Snakefile:
Job Missing files after 5 seconds:
results/simple_model/report
This might be due to filesystem latency. If that is the case, consider to increase the wait time with --latency-wait.
Job id: 0 completed successfully, but some output files are missing. 0
  File "c:\users\meinm\anaconda3\envs\rli\lib\site-packages\snakemake\executors\__init__.py", line 584, in handle_job_success
  File "c:\users\meinm\anaconda3\envs\rli\lib\site-packages\snakemake\executors\__init__.py", line 252, in handle_job_success
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: C:\Users\meinm\Documents\Git\oemof-B3\.snakemake\log\2021-05-27T133129.088985.snakemake.log
```

jnnr:
With the last commit, I could fix the remaining errors of the rule. Please check if it works on your system as well, @juliusmeier!

jnnr:
Remaining problem: The plots' filenames contain a space. Please make sure to always use underscores. Apart from that, it should work now.

juliusmeier:
It is now possible to create a .pdf report. 
In order to create a .html report, the plot images need to be treated differently. There is another option required to use the option `--resource-path=` as described here: https://groups.google.com/g/pandoc-discuss/c/RwQLMOH06hE/m/LbtTjXIdBQAJ

juliusmeier:
Now, a report file in .html format is also produced.
But the plots are tiny. I don't know how to handle images in html. I included report.css to set the style of report.html and I thought it would be possible to make some general settings for images there, too. But I did not find a way to do so.
Any suggestions how to display the plots better in report.html? @jnnr

nesnoj:
Hey, I'm happy to see that you're successful in using snakemake :).
In the djagora project I also implemented the generation of report (snakemake report format) and different separate html files. I used plotly to make the html files interactive which is quite nice. not sure if this is of interest for you but if yes, I could share some snippets (unfortunately, the repo is not public atm)...

jnnr:
> In the djagora project I also implemented the generation of report
Thanks for visiting us, @nesnoj! I would be very curious to see some snippets and maybe some of the resulting htmls!

jnnr:
> Now, a report file in .html format is also produced.

Thanks for this, @juliusmeier! I had a look and I agree, this does not look nice. Could not find a solution now, maybe we can go back to .png - this should render more nicely.

nesnoj:
> Thanks for visiting us, @nesnoj! I would be very curious to see some snippets and maybe some of the resulting htmls!

You should have received a NC link @jnnr, unfortunately @juliusmeier doesn't seem to have a NC account and I do not want to share it publicly, sorry.
I'm quite interested in how you integrated the reporting, what about spending some minutes of a lunch some day?

PR #32, opened 2021-05-05
Features/plot results timeseries

juliusmeier:
Create a plotting script and a snakemake rule which plots the dispatch from postprocessed data. The plots are supposed to be saved in a new directory called plotted.
As an example a sommer month and a winter month are plotted.


juliusmeier:
How can I avoid the error during the linting with black?

juliusmeier:
The snakemake rule plot_all_examples and the script plotting.py are now able to plot the dispatch of all csv-files which have electricity in their file name in the directory postprocessed. The legend of the plot is displayed below. The plots are saved as .pdf in a new directory called plotted in each scenario results.

jnnr:
> How can I avoid the error during the linting with black? @jnnr
> We talked about it once but I am not sure anymore about how to handle it.

You can run black <path-to-file> to let black reformat it. If that worked, you can commit and push the changes.

jnnr (approved):
Tested and like the results! Beautiful plots.

Could rename the script to "plot_dispatch" to parallel the name of the rule. Also, a docstring header as the one in "prepare_conv_pp.py" would be nice. No need to write tons of text, but a little documentation always helps!

Do you get this warning here as well and if yes, can you take care of it?

FutureWarning: Using an implicitly registered datetime converter for a matplotlib plotting method. The converter was registered by pandas on import. Future versions of pandas will require you to explicitly register matplotlib converters.

To register the converters:
        >>> from pandas.plotting import register_matplotlib_converters
        >>> register_matplotlib_converters()
  warnings.warn(msg, FutureWarning)

juliusmeier:
> Do you get this warning here as well and if yes, can you take care of it?
> 
> FutureWarning: Using an implicitly registered datetime converter for a matplotlib plotting method. The converter was registered by pandas on import. Future versions of pandas will require you to explicitly register matplotlib converters.
> 
> To register the converters:
> >>> from pandas.plotting import register_matplotlib_converters
> >>> register_matplotlib_converters()
> warnings.warn(msg, FutureWarning)

I do not get this warning but I think I fixed it. Could you test it?
I added a line with the registration of the matplotlib converter to pandas in the plots.py script in oemoflex: 
https://github.com/rl-institut/oemoflex/pull/7/commits/999073e2fa1c62111cdc457859bfa754f2d23052

jnnr (changes requested):
Thanks for your effort! Only small things:

* Please add a header to plot_dispatch.py, as in prepare_conv_pp.py, listing inputs and outputs and one or two sentences on what the scripts does.
* Test my last commit which drops the snake_copy.py script and uses python in the Snakefile directly.
* See if you like the changes I made to the script. I renamed loop variables and separated the definition of the selection of time and carrier from the loops.

juliusmeier:
> Thanks for your effort! Only small things:
> 
> * Please add a header to plot_dispatch.py, as in prepare_conv_pp.py, listing inputs and outputs and one or two sentences on what the scripts does.
> * Test my last commit which drops the snake_copy.py script and uses python in the Snakefile directly.
> * See if you like the changes I made to the script. I renamed loop variables and separated the definition of the selection of time and carrier from the loops.

* docstring header: Done
* copy command directly in snakefile: Works. So much cleaner without having to use the copy.py script. Thanks!
* Thanks for the improvements in the script. It is much better now.

jnnr (approved):
Thanks again! Approve. Can this https://github.com/rl-institut/oemoflex/pull/7 be merged as it is connected?

jnnr (approved):

PR #34, opened 2021-05-31
Update example scenario documentation

jnnr:
This PR renames the example scenarios in the docs and adds some more text.

TODO
* [x] Update the capacities to match the current version of NEP in the link (done in https://github.com/rl-institut/oemof-B3/pull/11)
* [x] Create a proper citation of the NEP document stating the version of the draft and an access date.

### DIFF
--- a/docs/bibliography.bib
+++ b/docs/bibliography.bib
@@ -11,10 +11,10 @@ volume = {22},
 year = {2018}
 }
 
-@online{NEP2021_Entwurf_1,
+@misc{NEP2021_Entwurf_1,
 author      = {Übertragungsnetzbetreiber CC-BY-4.0},
 title       = {Netzentwicklungsplan Strom 2035 (2021), erster Entwurf},
 year        = {2021},
 url         = {https://www.netzentwicklungsplan.de/sites/default/files/paragraphs-files/NEP_2035_V2021_1_Entwurf_Teil1.pdf},
-urldate     = {2021-06-08}
+note        = {accessed 2021-06-08}
 }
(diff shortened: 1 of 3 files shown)


jnnr:
Thanks for 9f12b6a, @juliusmeier! Unfortunately, I get an error. sphinxcontrib-bibtex does not seem to recognize the item type "online". I tried to change it to "report", but got the same result.

juliusmeier:
> Unfortunately, I get an error. sphinxcontrib-bibtex does not seem to recognize the item type "online". I tried to change it to "report", but got the same result.

On my computer, it works. But I will have a look.

jnnr:
Thanks for updating. Now it works for me as well. But the citation itself does not look

jnnr:
> Thanks for updating. Now it works for me as well.
> 
> The citation itself does not look proper yet. The license (CC-BY 4.0) seems misplaced. Maybe you can even drop the author (Übertragungsnetzbetreiber).

Seems like this is the way the document is supposed to be cited. In that case, this is ready to be merged, right? @juliusmeier

juliusmeier:
> Seems like this is the way the document is supposed to be cited. In that case, this is ready to be merged, right? @juliusmeier

Yes, I merge it.

jnnr:
From my side, this seems finished and ready to merge. Any objections, @juliusmeier?

juliusmeier:
> From my side, this seems finished and ready to merge. Any objections, @juliusmeier?

No, can be merged. I will do it now.

PR #39, opened 2021-06-24
Fix report rule in snakemake file

juliusmeier:
I could not run the report rule. I discovered that it must be connected with the use of quotation marks. Apparently the attribute title of the metadata demands "..." instead of '...'
For the shell command I replaced the "..." with '...' so that there is no confusion between the shell command and the title string.
Now the rule works for me (Windows machine) and I suggest to adapt this change.


jnnr (approved):
Tested on Ubuntu, works.

PR #40, opened 2021-06-24
Sketch for report with interactive elements

jnnr:
This adds a draft for a report with interactive plotly figures included as html. 

To allow to include html that do not contain the plotly library, a javascript link is included in the header of report_interactive.md. This should allow to save the plotly figures without including the plotly librrary, which reduces file size. 

When I tested it that way, the figures where not shown in the report, though.


jnnr:
Now, two interactive plots are included in the interactive report. The plotly library is excluded from the plots. The idea is that it can be loaded in the header of the html report. This does not work yet.

jnnr:
Tested it and it works. Setting the script via header-include seems unnecessary, as the files produced by plotly already include the link to the script source.

jnnr:
TODO:

* [ ] Optimize the appearance. At the moment, the plots exceed the margins. Do we need the margins?

juliusmeier:
In order to make the interactive report work, the plotly html-file is now set with the parameters `include_plotlyjs = "cdn" ` and `full_html=Flase` according to https://plotly.github.io/plotly.py-docs/generated/plotly.io.write_html.html
The files are much smaller now because a script tag containing the plotly.js source code (~3MB) is not included in the output anymore.
It did not work with `include_plotlyjs = False ` and `full_html=Flase` and loading plotlyjs in the html-file.

jnnr:
> In order to make the interactive report work, the plotly html-file is now set with the parameters `include_plotlyjs = "cdn" ` and `full_html=Flase` according to https://plotly.github.io/plotly.py-docs/generated/plotly.io.write_html.html
> The files are much smaller now because a script tag containing the plotly.js source code (~3MB) is not included in the output anymore.
> It did not work with `include_plotlyjs = False ` and `full_html=Flase` and loading plotlyjs in the html-file.

Can you add an inline comment in scripts/plot_dispatch.py at the respective line which explains this (can be shorter than this explanation here).

juliusmeier:
Including plotlyjs in a html-file of a single plotly plot makes it ca. 3 MB bigger. (here: 6 MB with it and 3 MB without it)
The interactive report is 26 MB with it vs. 10 MB without it using CDN.

jnnr:
> Including plotlyjs in a html-file of a single plotly plot makes it ca. 3 MB bigger. (here: 6 MB with it and 3 MB without it)
> The interactive report is 26 MB with it vs. 10 MB without it using CDN.

It is funny that the 2 plots which are the main ingredients are both 3MB but the interactive report in the end is 10MB. Any idea why?

jnnr:
@juliusmeier: This is almost ready. Please have a look at the script header of `plot_dispatch.py` and make sure that it is up-to-date. I guess that the .html output is not listed yet.

juliusmeier:
> It is funny that the 2 plots which are the main ingredients are both 3MB but the interactive report in the end is 10MB. Any idea why?

Unfortunately, I have no idea why.

PR #43, opened 2021-07-01
Set up toy scenarios

jnnr:
This PR introduces a basic scenario with components according to oemof-B3's model structure (see the sketch here https://oemof-b3.readthedocs.io/en/latest/overview.html). 

Extra components are defined.

TODO 
* [x] define in scenario yaml which data is to be used,
* [x] solve problem: EnergyDatapackage.create_default() gets unexpected arguments from scenario_yaml 
* [x] solve problem: build_datapackage and parametrize have the same output directory. Join scripts?
* [x] parametrize scalars and timeseries.


jnnr:
Some thoughts regarding scenario yaml and parametrization:

- Should the detailled file paths or filters be set here?
- Or does is there always one scenario file .csv for each scenario implied?
- At the same time, file paths would be a task for snakemake.
- Not everything is run the same for every scenario, some are investment runs, other not, etc.
- Some scenario data are raw, some are resources.

PR #45, opened 2021-07-08
Adapt to the change to own mapping in oemoflex plotting functions

juliusmeier:
The structure of the oemoflex plotting functions is changed to allow own mapping. Therefore some changes have to be made in plot_dispatch.py


jnnr (approved):
Looks good! Just needs  resolving of the merge conflict with dev. Then you can merge.

PR #46, opened 2021-07-12
Plot scalars of joined scenarios

juliusmeier:
The resulting scalar data of the scenarios are plotted together in one plot to better compare the data.

### DIFF
--- a/scripts/plot_joined_scalars.py
+++ b/scripts/plot_joined_scalars.py
@@ -1,0 +1,147 @@
+import pandas as pd
+import matplotlib.pyplot as plt
+import numpy as np
+import os
+from collections import OrderedDict
+import oemoflex.tools.plots as plots
+
+q = "capacity"
+w = "flow_in_electricity_from_bus"
+e = "flow_in_electricity_to_bus"
+r = "flow_out_electricity_from_bus"
+t = "flow_out_electricity_to_bus"
+z = "transmission_losses"
+
+u = "flow_in_biomass"
+i = "flow_out_electricity"
+o = "invest_out_electricity"
+p = "summed_variable_costs_in_biomass"
+a = "summed_variable_costs_out_electricity"
+s = "flow_in_electricity"
+d = "invest"
+f = "invest_in_electricity"
+g = "storage_capacity"
+h = "storage_losses"
+j = "flow_in_ch4"
+k = "summed_variable_costs_in_ch4"
+
+var_name = q
+
+names_dict = {"BE-biomass-st":"Biomass","BB-biomass-st":"Biomass",
+              "BE-electricity-curtailment":"El. curtailment","BB-electricity-curtailment":"El. Curtailment",
+              "BE-electricity-demand":"El demand","BB-electricity-demand":"El. Demand",
+              "BE-electricity-liion-battery":"Liion Battery","BB-electricity-liion-battery":"Liion Battery",
+              "BE-electricity-shortage":"El. shortage","BB-electricity-shortage":"El. Shortage",
+              "BE-BB-electricity-transmission":"El. Transmission",
+              "BE-ch4-gt": "CH4", "BB-ch4-gt": "CH4",
+              "BE-solar-pv":"PV","BB-solar-pv":"PV",
+              "BE-wind-onshore":"Wind on","BB-wind-onshore":"Wind on",
+              }
+
+unit_dict = {"capacity":"W"}
+
+# load colors dict which also determines the ordering
+colors_csv = pd.read_csv("colors.csv", header=[0], index_col=[0])
+colors_csv = colors_csv.T
+colors_odict = OrderedDict()
+for i in colors_csv.columns:
+    colors_odict[i] = colors_csv.loc["Color", i]
+
+# import data
+data_path = r"C:\Users\meinm\Documents\Git\oemof-B3\results\joined_scenarios\examples\scalars.csv"
+scalars = pd.read_csv(data_path)
+
+# select data with chosen var_name
+selected_scalar_data = scalars[scalars['var_name'] == var_name]
+
+
+def prepare_scalar_data(df, colors_odict, names_dict):
+    # rename
+    df = df.copy()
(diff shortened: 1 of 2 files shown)


jnnr (commented):
Thanks for your work!

In general, I was asking myself if it is possible to separate the bar plot and the correct setting of the "multi-level" ticks. This would improve the readability of the code and would allow to plot scalars with single-level ticks as well. As you are using multiple ax'es though, this would not be so straight-forward.

Can you please make the script ready to be included in the snakemake pipeline as well? I had to adapt the paths to let it run for testing.

scripts/plot_joined_scalars.py
```diff
@@ -0,0 +1,147 @@
...
+              }
+
+unit_dict = {"capacity":"W"}
+
+# load colors dict which also determines the ordering
+colors_csv = pd.read_csv("colors.csv", header=[0], index_col=[0])
+colors_csv = colors_csv.T
+colors_odict = OrderedDict()
+for i in colors_csv.columns:
+    colors_odict[i] = colors_csv.loc["Color", i]
```
jnnr, line 48: Here, you could use the colors defined in oemoflex.

```suggestion
from oemoflex.tools.plots import colors_odict
```

scripts/plot_joined_scalars.py
```diff
@@ -0,0 +1,147 @@
...
+names_dict = {"BE-biomass-st":"Biomass","BB-biomass-st":"Biomass",
+              "BE-electricity-curtailment":"El. curtailment","BB-electricity-curtailment":"El. Curtailment",
+              "BE-electricity-demand":"El demand","BB-electricity-demand":"El. Demand",
+              "BE-electricity-liion-battery":"Liion Battery","BB-electricity-liion-battery":"Liion Battery",
+              "BE-electricity-shortage":"El. shortage","BB-electricity-shortage":"El. Shortage",
+              "BE-BB-electricity-transmission":"El. Transmission",
+              "BE-ch4-gt": "CH4", "BB-ch4-gt": "CH4",
+              "BE-solar-pv":"PV","BB-solar-pv":"PV",
+              "BE-wind-onshore":"Wind on","BB-wind-onshore":"Wind on",
+              }
```
jnnr, line 39: Here, we should use the labels from oemoflex.

scripts/plot_joined_scalars.py
```diff
@@ -0,0 +1,147 @@
...
+    plt.tight_layout()
+    plt.savefig(var_name + ".png",bbox_inches='tight')
+
+
+# prepare data
+print(selected_scalar_data)
+prepared_scalar_data, colors = prepare_scalar_data(df=selected_scalar_data, colors_odict=colors_odict, names_dict=names_dict)
+print(prepared_scalar_data)
+# plot data
+plot_scalars(df = prepared_scalar_data, colors=colors, unit_dict=unit_dict, var_name=var_name)
```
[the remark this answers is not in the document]
jnnr: Now I see that this is difficult because what you create is a gridspec.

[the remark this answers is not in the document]
juliusmeier: Is there an alternative to gridspec?

scripts/plot_joined_scalars.py
```diff
@@ -0,0 +1,147 @@
...
+        if i not in df_pivot.columns:
+            continue
+        colors.append(colors_odict[i])
+        concrete_order.append(i)
+    df_pivot = df_pivot[concrete_order]
+
+    return df_pivot, colors
+
+
+def plot_scalars(df, colors, unit_dict, var_name):
```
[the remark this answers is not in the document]
juliusmeier: The underlying problem I want to solve is that the unit changes depending on the var_name. I thought of creating a dict with all var_names and their corresponding units.

jnnr (changes requested):
Thanks for your work. Tested it with the new label and color mapping in oemoflex. I encountered the problem that the labels would not match as the scalar results already have single string labels, not tuples. I adapted the functions of oemoflex.tools.plots such that it can map single strings as well (https://github.com/rl-institut/oemoflex/pull/32). After that, I still get errors (KeyError: "['Shortage'] not found in axis"
and AttributeError: 'SubplotSpec' object has no attribute 'is_first_col') Please make sure that your branch works with that branch.

juliusmeier:
Now everything works on my computer with https://github.com/rl-institut/oemoflex/pull/32

jnnr:
I get this error:
```
  File "scripts/plot_joined_scalars.py", line 146, in plot_scalars
    if ax.get_subplotspec().is_first_col():
AttributeError: 'SubplotSpec' object has no attribute 'is_first_col'
```

Which matplotlib version are you using?

juliusmeier:
> Which matplotlib version are you using? I use 3.3.3

I use `matplotlib-base           3.4.2            py39h581301d_0    conda-forge`

jnnr:
I have problems to install a matplotlib version > 3.3.4 at the moment. Tested the script without the call of `is_first_col`, which worked.

Snakefile
```diff
@@ -4,9 +4,12 @@ examples = [
     'more_renewables_less_fossil'
 ]
 
-scenarios = ["toy-scenario", "toy-scenario-2"]
+scenario_list_example = ['examples']
```
jnnr, line 7: With this change, snakemake does not find the variable scenarios any more, which is needed below.

scripts/plot_joined_scalars.py
```diff
@@ -0,0 +1,176 @@
...
+-------------
+The result scalars of all scenarios are plotted in a single plot.
+"""
+import sys
+import os
+import pandas as pd
+import matplotlib.pyplot as plt
+import numpy as np
+import oemoflex.tools.plots as plots
+from oemoflex.tools.plots import colors_odict
```
jnnr, line 24: Please import the labels and colors of oemof_b3 here and pass them to the plotting functions below.

`from oemof_b3 import labels_dict, colors_odict`

PR #47, opened 2021-07-12
Stack/Unstack timeseries to/from timeseries data template

jnnr:
In #41, a model-wide data format for time series has been introduced. It represents timeseries in a form where one series resides in one row. To be able to process this format in the model, we need functions that transform to and from that format, e.g.
```
timeindex        A B C
2012-01-01 00:00 1 2 1
2012-01-01 01:00 1 2 1
2012-01-01 02:00 1 2 1
```
from/to
```
var_name timeindex_start timeindex_end timeindex_resolution series 
A 2012-01-01 00:00 2012-01-01 02:00 'H' [1,1,1]
B 2012-01-01 00:00 2012-01-01 02:00 'H' [2,2,2]
C 2012-01-01 00:00 2012-01-01 02:00 'H' [1,1,1]
```


MaGering:
Hi @jnnr, I've updated the function for stacking and unstacking time series in this PR and also the tests.

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+
+
+def check_consistency_timeindex(df, index, name):
+    """
+    This function assert that values of a column in a stacked DataFrame are same
+    for all time steps
+
+    Parameters
+    ----------
```
jnnr, line 11: Missing parameter: df

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,121 @@
...
+    index : string
+        Index of values to be checked in the DataFrame
+    name : string
```
jnnr, line 14: 'name' is only used for the error message, no other function. Could we drop it?

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,121 @@
...
+            "'%Y-%m-%d %H:%M:%S'."
+        )
+
+    # Assert that frequency match for all time steps
+    if pd.infer_freq(_df.index) is None:
+        raise TypeError(
+            "No frequency of your provided data could be detected."
+            "Please provide a DataFrame with a specific frequency (eg. 'H' or 'T')."
+        )
+    else:
```
jnnr, line 55: You don't need this "else".

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,121 @@
...
+        "timeindex_stop",
+        "timeindex_resolution",
+        "series",
+    ]
+
+    df_stacked = pd.DataFrame(columns=df_stacked_cols)
+
+    for column in df.columns:
+        var_name = column
+        timeindex_start = df[column].index.values[0]
```
jnnr, line 79: The index should be the same for all columns. Could just do this outside the loop:

```suggestion
        timeindex_start = df.index.values[0]
```

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,121 @@
...
+        timeindex_stop = df[column].index.values[len(df[column]) - 1]
```
jnnr, line 80:
```suggestion
        timeindex_stop = df[column].index.values[-1]
```

jnnr (commented):
Thanks for your work,  @MaGering! I added some comments. Have a look if they are helpful. Please also add docstrings to the stack/unstack_timeseries functions.

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/8b0ea2db7abcb99128185834960a17a2a75efb79.

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/8b0ea2db7abcb99128185834960a17a2a75efb79.

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/8b0ea2db7abcb99128185834960a17a2a75efb79.

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/8b0ea2db7abcb99128185834960a17a2a75efb79.

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/8b0ea2db7abcb99128185834960a17a2a75efb79.

MaGering:
@jnnr with commit https://github.com/rl-institut/oemof-B3/pull/47/commits/f81058d5d87aeebc80267b4018fde5d95c7996a0 I introduce a workaround to a problem of index naming.

Background: If a DataFrame, which has a name for the index, is stacked and then unstacked again, this name is lost, because it was not processed during stacking so far.

I have now adapted the functions so that the name in the stacked DataFrame is stored in the index name of the column "timeindex_start". Unstacking a stacked DataFrame the name is extracted from that column's index name.

tests/test_data_processing.py
```diff
@@ -0,0 +1,70 @@
...
+    pd.testing.assert_frame_equal(ts_column_wise_again, ts_column_wise)
+
+    # In case the test does not error out it is None. Hence a passed test results
+    # to None
+    assert pd.testing.assert_frame_equal(ts_column_wise_again, ts_column_wise) is None
+    with pytest.raises(AssertionError):
+        pd.testing.assert_frame_equal(ts_column_wise_again, ts_column_wise_different)
+
+
+def test_stack_unstack_on_examples():
```
jnnr, line 51: This relies on the example being run already, which is a bit dodgy.

jnnr: Actually, that is not true: The data does not have to be created, it is already there. Still, I suggest that you rather take a stripped-down smaller version (i.e. just 3 timesteps) of the data used here, save it under test/_files/, push it and use this for the test.

MaGering: Done with commits https://github.com/rl-institut/oemof-B3/pull/47/commits/cf045753eca2eef4ffe181dc3edb915dd2dc9281 and https://github.com/rl-institut/oemof-B3/pull/47/commits/ab3292d73a7b78e22fa66f1f5888c9bd64eec828.

oemof_b3/tools/data_processing.py
```diff
@@ -0,0 +1,165 @@
...
+            series,
+        ]
+
+        dict_stacked_column = dict(zip(df_stacked_cols, column_data))
+        df_stacked_column = pd.DataFrame(data=dict_stacked_column)
+        df_stacked = df_stacked.append(df_stacked_column, ignore_index=True)
+
+    # Save name of the index in the unstacked DataFrame as name of the index of "timeindex_start"
+    # column of stacked DataFrame, so that it can be extracted from it when unstacked again.
+    df_stacked["timeindex_start"].index.name = _df.index.name
```
MaGering, line 120: @jnnr, do you see any difficulty implementing it like this for [this purpose](https://github.com/rl-institut/oemof-B3/pull/47#issuecomment-886609489)?

jnnr: I am not sure if this works. It seems that you assign the former index' name to the index name of df_stacked["timeindex_start"], which is a Series (or maybe a DataFrame with only one column). If you stack several timeseries with different index names, this does not work (not sure if this will ever happen, though).

MaGering: > If you stack several timeseries with different index names, this does not work (not sure if this will ever happen, though).

But the function can't stack several time series so far. Should we hence leave it like this for now or name it `timeindex` as default?

jnnr: It cannot, but if you could concat two stacked frames. I do not think this is too important, though. What we agreed on was the template, which is stacked. If you unstack from the template using your function, your timeindex will be called timeindex. This is reasonable, as the timeindex is specified in the stacked format by "timeindex_start" etc.

MaGering: Ok! Is this PR ready to be merged from your side?

jnnr: I am fine with merging :+1:

PR #50, opened 2021-07-14
Features/plot scalar resources

jnnr:
This introduces a script and a snakemake rule for plotting scalar resources.


scripts/plot_scalar_resources.py
```diff
@@ -33,33 +35,76 @@ def load_scalars(path):
...
+        "oil": "Oil",
+        "lignite": "Lignite",
+        "other": "Other",
+    }
+
+    # User input
+    var_name = "capacity_net_el"
+    ylabel = "Leistung [MW]"
+
+    def prepare_conv_pp_scalars(var_name, carrier_dict=carrier_dict):
```
jnnr, line 53: I would suggest move the definitions and functions to the to, above `if __name__ == '__main__`:.

scripts/plot_scalar_resources.py
```diff
@@ -33,33 +35,76 @@ def load_scalars(path):
...
+        # select var_name to be plotted
+        selected_df = scalars[scalars["var_name"] == var_name].copy()
+
+        # CONV_PP specific
```
jnnr, line 57: The script's name suggests that it is generic. How can we handle this?

scripts/plot_scalar_resources.py
```diff
@@ -0,0 +1,110 @@
...
+            columns="carrier",
+            values="var_value",
+        )
+
+        # colors
+        color_keys = df_pivot.columns
+
+        return df_pivot, color_keys
+
+    def plot_scalar_resources(df, color_keys):
```
jnnr, line 80: The plotting function could be more generic `plot_scalars` and moved to oemof-B3/tools/plots.py so that it can be reused.

scripts/plot_scalar_resources.py
```diff
@@ -0,0 +1,137 @@
...
+    # create the directory plotted where all plots are saved
+    if not os.path.exists(target_dir):
+        os.makedirs(target_dir)
+
+    def load_scalars(path):
+        df = pd.read_csv(path, sep=",", index_col=0)
+        return df
+
+    # Load scalar data
+    scalars = load_scalars(resources)
```
jnnr, line 42: This can be replace with the now merged load_b3_scalars.

scripts/plot_scalar_resources.py
```diff
@@ -0,0 +1,137 @@
...
+        # apply pivot table
+        df_pivot = pd.pivot_table(
+            selected_df_agg,
+            index=["region"],
+            columns="carrier",
+            values="var_value",
+        )
+
+        # colors
+        color_keys = df_pivot.columns
```
jnnr, line 91: Would be good to make clear where the translation from English to German happens, so that it can be switched on and off easily.

scripts/plot_scalar_resources.py
```diff
@@ -0,0 +1,137 @@
...
+        "CH4": "Erdgas",
+        "Hard coal": "Steinkohle",
+        "Oil": "Öl",
+        "Lignite": "Braunkohle",
+        "Other": "Andere",
+    }
+
+    # User input
+    var_name = "capacity_net_el"
+    conv_number = 1000 * 1000
```
jnnr, line 66: This is a bit obscure. An inline comment would help (converting Megawatt to Watt, I guess?). Also, should just write 1e6.

jnnr (commented):
Thanks for your work! This is almost finished. Just a few steps cleaning up to help our future selves.

Issue #48, opened 2021-07-14
Conv_pp does not follow the template data structure


jnnr:
#41 introduced a template data structure for all resources preprared in the model as well as all raw data that we compile ourselves.

conv_pp.csv is not produced according to this format yet. Please adapt the preparation such that it does.

juliusmeier:
Do we use CSV with commas (,) or semicolons (;) as separation? Or does it not matter?

juliusmeier:
What is the meaning of `id_scla` ?

PR #51, opened 2021-07-15
Adapt data structure of conventional pp to scalar data template

juliusmeier:
The prepared data of conventional power plant data is adapted to the template of scalar data.
Closes issue #48

### DIFF
--- a/scripts/prepare_conv_pp.py
+++ b/scripts/prepare_conv_pp.py
@@ -98,12 +98,18 @@ if __name__ == "__main__":
     b3_agg.loc[
         ("Oder-Spree", "Other fuels", "Steam turbine", "yes"), "efficiency_estimate"
     ] = b3_agg.loc[
-        b3_agg.index.get_level_values("energy_source") == "Waste", "efficiency_estimate"
+        b3_agg.index.get_level_values("energy_source") == "Other fuels", "efficiency_estimate"
     ].mean()
 
     b3_agg.reset_index(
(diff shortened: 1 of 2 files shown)


jnnr (changes requested):
Thanks for your work! I produced the resource and found the following:

* name, type should be None
* unit should be None for chp and efficiency
* map carrier and tech from OPSD terminology to oemoflex's carriers and techs
* reference: add the items to a bibtex (references_input_data.bib, separate from the bibtex that is already included in the docs) and state the citation key here.

juliusmeier:
- [ ] map carrier and tech from OPSD terminology to oemoflex's carriers and techs

juliusmeier (commented):
Done

juliusmeier (commented):
Done

scripts/prepare_conv_pp.py
```diff
@@ -45,7 +48,7 @@
 
     regions_nuts3_de = geo.load_regions_file(in_path2)
 
-    b3_regions_yaml = open(in_path3, "r")
+    b3_regions_yaml = open(in_path3, "r", encoding="utf-8")
     b3_regions_content = yaml.load(b3_regions_yaml, Loader=yaml.FullLoader)
```
jnnr, line 52:
```suggestion
    with open(in_path3, "r", encoding="utf-8") as b3_regions_yaml:
        b3_regions_content = yaml.load(b3_regions_yaml, Loader=yaml.FullLoader)
```

PR #57, opened 2021-08-11
Features/process data

### LINKED ISSUE #49: Missing helper function for template data format
As it may used in many of the scripts, we should introduce functions that handle data which is in the template data format (introduced in #41).

In particular, we need implementations for these:

* df = load_scalars(path) 
* df = load_timeseries(path)
* save_scalars(df, path)
* save_timeseries(df, path)
* df_filtered = filter(df, key) (key could be e.g. scenario, region, carrier, tech, type, var_name)
* df_agg = aggregate(df, key) (key could be e.g region, carrier, tech,)

MaGering:
This PR resolves #49 by adding the following functions:

- df_header = get_optional_required_header(data_type)
- df = load_scalars(path)
- df = load_timeseries(path)
- save_df(df, path)
- df_filtered_by_value = df_filtered(df, key, values)
- df_agg_by_key = df_agg(df, key)

On this PR I also adapted other functions in `data_processing.py` and `test_data_processing.py`.


oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+        or list(df.columns) == required_header_without_reg
+    ) and list(df.columns) != header:
+        # ID in the form of numbering is added if "id_ts" is missing
+        if optional_header[0] not in df.columns:
+            df[optional_header[0]] = np.arange(0, len(df))
+
+        # The region is extracted out of "var_name"
+        if required_header[0] not in df.columns:
+            region = []
+            for row in np.arange(0, len(df)):
```
jnnr, line 275: Seems that the function relies on the data (Hard coded BE, BB). This should not be.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+
+    """
+    # Save scalars to csv file
+    df.to_csv(path, index=False)
+
+    # Print user info
+    print(f"User info: The scalars have been saved to: {path}.")
+
+
+def save_timeseries(df, path):
```
jnnr, line 326: Two functions that do the same, so why not just one?

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+
+    # Empty DataFrame, which will contain aggregated items
+    df_agg_by_key = pd.DataFrame(columns=list(df.columns))
+
+    # Get list of all items existing in the column passed with key
+    key_list = list(df[key])
+    key_list = list(dict.fromkeys(key_list))
+
+    # Begin aggregation of scalars
+    if scalars:
```
jnnr, line 521: This is a hard-to-review part starting here: Deeply nested loops, lots of values are set. I would need more time to understand this. Hard to maintain as well. We should try to drastically simplify this. Let's talk about it.

jnnr (commented):
Thanks for your work! Will be very helpful. There is some potential for simplification which we should use. I commented some smaller suggestions. The bigger things would need more time to digest. I can give you more feedback if I take some time.

oemof_b3/tools/data_processing.py
MaGering: Good point! :) Done with commit https://github.com/rl-institut/oemof-B3/pull/57/commits/17ac36d0c4bbffa7b884c219cc7ebf4518a75ef3.

SabineHaas:
@MaGering `load_scalars()` worked fine for my purpose :+1:

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+        )
+
+    # Check whether optional columns are missing
+    for optional in optional_header:
+        if optional not in df_header:
+            # ID in the form of numbering is added if "id_scal" is missing
+            if optional is optional_header[0]:
+                df[optional] = np.arange(0, len(df))
+            else:
+                newline = "\n"
```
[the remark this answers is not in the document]
MaGering: Unfortunately I couldn't use `\n` directly in a f-string. That's why I used the variable `newline` as a workaround.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+
+    # Interrupt if required columns are missing and print all affected columns
+    if len(missing_required) > 0:
+        raise KeyError(
+            f"The data in {filename} is missing the required column(s): {missing_required}"
+        )
+
+    # Check whether optional columns are missing
+    for optional in optional_header:
+        if optional not in df_header:
```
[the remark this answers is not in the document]
MaGering: Because it does not appear in the results, neither in the scalar nor in the time series. So I wrote the script in a way that the id is variable and this column (which is just a numbering of the DataFrame, isn't it?) is added in case it is missing.

oemof_b3/tools/data_processing.py
MaGering: But do we need other regions than these three?

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+    -------
+    df : pd.DataFrame
+        DataFrame with loaded scalars
+
+    """
+    # Get header of scalars
+    scalars_header = get_optional_required_header("scalars")
+    header = scalars_header[0]
+    optional_header = scalars_header[1]
+    required_header = scalars_header[2]
```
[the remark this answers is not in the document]
jnnr: This change would make the code shorter and easier to read.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,761 @@
...
+    for optional in optional_header:
+        if optional not in df_header:
+            # ID in the form of numbering is added if "id_scal" is missing
+            if optional is optional_header[0]:
+                df[optional] = np.arange(0, len(df))
+            else:
+                newline = "\n"
+                # For every other optional column name, an empty array is added with the name as
+                # header - A user info is printed
+                df[optional] = [np.nan] * len(df["var_value"])
```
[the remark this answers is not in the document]
jnnr: Did you try?

oemof_b3/tools/data_processing.py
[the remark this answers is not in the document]
jnnr: Ok, I see.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,734 @@
...
+
+    """
+    # Save scalars to csv file
+    df.to_csv(path, index=False)
+
+    # Print user info
+    print(f"User info: The DataFrame has been saved to: {path}.")
+
+
+def df_filtered(df, key, values):
```
jnnr, line 319:
```suggestion
def filter_df(df, key, values):
```

jnnr: I would name the function by its action, so using the function could look like `df_filtered = filter_df(df)`

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,734 @@
...
+    -------
+    df : pd.DataFrame
+        DataFrame with loaded time series
+
+    """
+    # Get header of time series
+    timeseries_header = get_optional_required_header("timeseries")
+    header = timeseries_header[0]
+    optional_header = timeseries_header[1]
+    required_header = timeseries_header[2]
```
jnnr, line 167:
```suggestion
    header, optional_header, required_header = get_optional_required_header("timeseries")
```

jnnr: This appears a couple of times.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,734 @@
...
+    df = pd.read_csv(path)
+
+    # Save header of DataFrame to variable
+    df_header = list(df.columns)
+
+    # Check whether required columns are missing in the DataFrame
+    missing_required = list(set(required_header).difference(set(df_header)))
+
+    # Interrupt if required columns are missing and print all affected columns
+    if len(missing_required) > 0:
```
jnnr, line 117:
```suggestion
    if missing_required:
```

oemof_b3/tools/data_processing.py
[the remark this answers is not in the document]
jnnr: I made a suggestion above how to handle this.

oemof_b3/tools/data_processing.py
```diff
@@ -1,7 +1,734 @@
...
+    """
+    # Get header of scalars
+    header, optional_header, required_header = get_optional_required_header("scalars")
+
+    # Get file name
+    filename = os.path.splitext(path)[0]
+
+    # Read data
+    df = pd.read_csv(path)
+
```
jnnr, line 109:
```suggestion
    # Set index
    if "id_scal" in df:
        df.set_index("id_scal", inplace=True)
        
    else:
        df.index.name = "id_scal"
```

tests/test_data_processing.py
```diff
@@ -26,14 +38,512 @@
...
+    1. scalars and
+    2. time series
+    are returned correctly
+
+    In a third test it is checked whether a ValueError
+    is raised if an invalid string is passed with
+    data_type
+
+    """
+    # 1. Test
```
jnnr, line 53: Please do not test 3 things in one test function, but rather one thing  in one test. If there are variables you would like to share among test functions, you can create test classes with the methods testing one thing at a time.

oemof_b3/tools/data_processing.py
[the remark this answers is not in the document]
MaGering: Unfortunately, I do not understand how this makes a difference to my variant. In both cases `id_scal` is added when it does not exist. If I set `id_scal` as index, I have to adjust the tests, because then it doesn't show up there as column anymore. Also saving the DataFrame I have to unindex `id_scal` so that this column is saved with the respective header don't I? Hence I see a lot of extra work with the index variant.

tests/test_data_processing.py
```diff
@@ -26,14 +38,512 @@
...
+    df = load_scalars(path_file_scalars)
+
+    # Save read scalars
+    save_df(df, path_file_saved)
+
+    # Load the saved scalars
+    df_saved = load_scalars(path_file_saved)
+
+    # 1. Test
+    assert os.path.exists(path_file_saved) == 1
```
jnnr, line 275: This seems obsolete as the test would fail in the previous line if the file would not exist.

tests/test_data_processing.py
```diff
@@ -26,14 +38,512 @@
...
+    )
+
+    # Read scalars
+    df = load_scalars(path_file_scalars)
+
+    # Aggregate by region
+    df_aggregated_by_region = df_agg(df, "region")
+
+    # Add expected results of aggregation
+    df_aggregated_expected = pd.DataFrame(
```
jnnr, line 397: Please save this as a .csv under _files.

tests/test_data_processing.py
```diff
@@ -26,14 +38,512 @@
...
+        }
+    )
+
+    # Check if results of aggregation equal the expected ones
+    pd.testing.assert_frame_equal(df_aggregated_by_region, df_aggregated_expected)
+
+    # Aggregate by carrier
+    df_aggregated_by_carrier = df_agg(df, "carrier")
+
+    assert list(df_aggregated_by_carrier["var_value"].values) == [
```
jnnr, line 438: Here as well, it would be more illustrative and easier to read if the expected file would reside in _files

tests/test_data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/57/commits/10429062c7180571172a6d83d4628f3da593b5ba.

tests/test_data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/57/commits/10429062c7180571172a6d83d4628f3da593b5ba.

tests/test_data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/57/commits/10429062c7180571172a6d83d4628f3da593b5ba.

tests/test_data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/57/commits/10429062c7180571172a6d83d4628f3da593b5ba.

MaGering:
I adapted the tests to your suggestion @jnnr. PIease consider that I couldn't match the dtypes in the aggregation and filter function of the origin and modified DataFrame. Hence these tests run without checking dtypes for now. I left a comment about this in the code.

PR #70, opened 2021-10-06
Add emission constraint

jnnr:
This PR introduces an emission constraint to the models. This is done by adding a constraint in the script 'optimize.py'. The emission factors have to be set to the respective flows by using `input_parameters` or `output_parameters` in the scalar input data.

TODO:

* [x] Decide where the script gets the information on the emission constraint from. From the scenario.yml, or from a separate csv that would be part of the preprocessed datapackage.
* :x: Parametrize the emission factors. Each carrier that has a nonzero emission factor should be included. --> The parametrization is a data issue and won't be solved in this PR.  (SH)
* :x: Test it. --> We are not convinced, yet, whether we will write tests for this. I have tested the constraint locally though. (SH)


SabineHaas:
I would like to provide the emission limit in scalars.csv. It is possible to add different emission limits (if we want to have scenarios with different limits) and filter by scenario-key. @jnnr I understood we are currently not supposed to add any entries into scalars csv files right?

Also emission factors I would add to scalars:
1) (positive) emisson factors I only see for H2 import, CO2 import and CH4 import (maybe biomass). H2 from elektrolysis has emission factor zero and CH4 from the methanisation storage does not have an emission factor in the model as it is already covered by CO2 import.
2) Then we have a negative emission factor for CH4 export. It is calculated from the emission factor of CO2 (how many units CO2 are needed for one unit of CH4?).
3) H2 export does not have a (negative) emission factor, as we only export H2 from electrolysis.*

*Actually this depends on the prices for H2 import and export.
If import is more expensive than export and we assume a constant price (which is the current aim, right @jnnr?) we would not earn money with buying H2, storing it and then selling it. In case we use price time series this could change; cheap H2 could be bought and stored and sold at a higher price at a later time of the year.

SabineHaas:
Notes:

- Currently there's two input files: `base-scenario_emissions_PR70.csv` (non-tested, including shortages) and `base-scenario_emissions_PR70_without_ch4_import.csv` (used to check the emission constraint) in raw directory
- Unfortunately emission constraint does not work as expected right now. However, simulation results look unexpected also without constraint. 
- I have noticed that `ch4-excess` has `output_parameters` as default in `oemoflex`, shouldn't that be `input_parameters`, as excess component does not have an output but input flow?

SabineHaas:
Open todos 

- [ ] add test with small system (less components) --> how to test snakemake rule

also possible in other PR

- [ ] fix: h2/ch4 excess/shortage components not shown in results

jnnr:
@SabineHaas: I guess we can delete this branch. If not, please protest.

PR #35, opened 2021-06-01
Renewable potential wind and pv: area and power potential

SabineHaas:
This PR includes the calculation of area and power potential of wind and pv.

PR #72, opened 2021-10-08
Add reference to header of scalars.csv 

SabineHaas:
Added "reference" to header of 'scalars.csv'  in directory `schema`.

This came up in #35. I got a ValueError because of "extra columns": `ValueError: There are extra columns ['reference']`
If "reference" stays in 'scalars.csv' this PR should be merged into `dev`.


SabineHaas:
Ahh I see, I have renamed 'reference' to 'source' in `scalars.csv` on our server now.

Issue #55, opened 2021-07-26
Feedin timeseries are missing


jnnr:
The model needs feed-in timeseries for renewables (hourly capacity factors).

* [x] PV from renewables ninja / pvlib bzw. feedinlib
* [x] Wind from renewables ninja / windpowerlib bzw. feedinlib
* [x] RoR

SabineHaas:
@jnnr 
Is there already a directory for the time series? Otherwise, I'd create `raw/time_series`?
For bringing the time series into the desired format we need to wait for #57 I assume

SabineHaas:
For now I have placed them into  'Daten/raw/time_series' on our server

jnnr:
> For now I have placed them into 'Daten/raw/time_series' on our server

That sounds reasonable, let's do it like this.

jnnr:
@SabineHaas, is this resolved and can we close it?

SabineHaas:
Yes, it was resolved with #68

PR #73, opened 2021-10-08
Feed-in time series wind and pv

SabineHaas:
Address #55

This PR adds a script and a rule for preparing wind and pv feed-in time series from renewables ninja for the years 2010 - 2019.

Note, that ROR time series could be included into the rule and script. @jnnr this might actually make sense as I assume that the single input files won't be changed frequently and run time is short, so better bundle the preparation in one rule to only receive one output file "time_series.csv".  --> decided to do so once this PR was merged

todos
- [x] check time zone --> UTC
- [x] tests
- [x] description
- [x] read `year` from scalars? maybe not scalars.. @jnnr what would be the right location for setting the year for the time series?

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -64,12 +64,11 @@ rule prepare_feedin:
     input:
         wind_feedin="raw/time_series/ninja_wind_country_DE_current_merra-2_nuts-2_corrected.csv",
         pv_feedin="raw/time_series/ninja_pv_country_DE_merra-2_nuts-2_corrected.csv",
-        time_series_template="oemof_b3/schema/timeseries.csv",
         script="scripts/prepare_feedin.py"
     output:
         "results/_resources/feedin_time_series.csv"
     shell:
-        "python {input.script} {input.wind_feedin} {input.pv_feedin} {input.time_series_template} {output}"
+        "python {input.script} {input.wind_feedin} {input.pv_feedin} {output}"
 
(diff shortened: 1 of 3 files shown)


SabineHaas:
> > Okay! +1 I would prepare 10 years only (2010 - 2019) - is this fine for you? RE ninja provides data from 1980, that seems too many unnecessary years to me and a too big file (~100 GB for wind). years 2010 - 2019 shrinks down to 25 GB.
> 
> That seems reasonable. Are the files really that big? The previous file for 2012 contained 4 rows and was less than 1MB.

Ah no, it's the raw files that are so big. e.g. `ninja_wind_country_DE_current_merra-2_nuts-2_corrected.csv`

SabineHaas:
> Thanks for the changes. Some remarks:
> 
>     * There is still a SettingWithCopyWarning you might want to fix.
> 

This copy warning is raised withing a pandas script: `/umas/lib/python3.8/site-packages/pandas/core/indexing.py:494: SettingWithCopyWarning:` 

>     * Please use `load_b3_timeseries` to avoid passing the template to the script.

There seems to be a bug. In `load_b3_timeseries` the column `id_ts` is set as index (by function `format_header`). Using then `format_header` to prepare my time series `format_header` tries to set the index again --> `KeyError: "None of ['id_ts'] are in the columns"`

PR #77, opened 2021-10-13
Features/annuisation

jnnr:
This PR introduces a script that calculates annuised investment cost from the raw scalar data on overnight cost, lifetime and wacc. It introduces new functions and a class to oemof_b3.tools.data_processing which makes it easier to bring the scalar data into a form that allows calculations and send the results back to the stacked scalar data format.


jnnr:
I updated the annuisation to include both `capacity_costs` and `storage_capacity_costs`. 

The script "create_empty_scalars.py" is cleaned up a bit. It is now simpler to create empty oemof_b3-scalars either in raw or resources (where annuisation already combined the cost, lifetime and wacc). 

Using the script, I created empty raw scalars for the basic scenario, set all values to zero and added a value for the wacc (on the fileserver now), annuised it with prepare scalars and fed it into the base-scenario defined in #69. It works, which means that the optimization script succeeds and passes the problem to the solver. 

This is one step towards feeding real data to the model. Next step is to check which real assumptions are missing in the raw scalars and complete them.

Issue #82, opened 2021-10-26
Docs of processing steps is not tidy


jnnr:
The scripts' header should be shown in the section [model pipeline of the docs](https://oemof-b3.readthedocs.io/en/latest/model_pipeline.html) in the respective chapters preprocessing, optimisation, postprocessing, plotting. This should help to understand the model pipeline. To make that happen, 

1. There has to be file in the respective folder docs/preprocessing rsp. docs/optimisation etc.
2. The folder has to be included in model pipeline (https://github.com/rl-institut/oemof-B3/blob/dev/docs/model_pipeline.rst)
3. The header has to exist and be formatted in a tidy way. In particular, the filenames of the inputs and outputs should be shown.

Tidy up documentation of processing steps by:

Make sure that all scripts are shown in the docs
- [x] build_datapackage.py
- [x] join_scenarios.py
- [x] optimize.py
- [x] plot_dispatch.py
- [x] plot_joined_scalars.py
- [x] plot_scalar_resources.py
- [x] postprocess.py
- [x] prepare_conv_pp.py
- [x] prepare_electricity_demand.py
- [x] prepare_feedin.py
- [x] prepare_heat_demand.py
- [x] prepare_re_potential.py
- [x] process_re_potential.py

Include the names of the files in the docs. This means that for each of the inputs and outputs to look up the filename in `Snakefile` and add it to the header. e.g. change this:

https://github.com/rl-institut/oemof-B3/blob/828d30423f5d0222d4ba05ec2f234bafec39ccf8/scripts/build_datapackage.py#L5-L6

to this:

```
scenario_specs : str
    ``scenarios/{scenario}.yml``: path of input file (.yml) containing scenario specifications
```

This will make the docs much more useful (and it seems ok to hard-code the filenames in the header, as they will not change too often).

- [x] build_datapackage.py
- [x] join_scenarios.py
- [x] optimize.py
- [x] plot_dispatch.py
- [x] plot_joined_scalars.py
- [x] plot_scalar_resources.py
- [x] postprocess.py
- [x] prepare_conv_pp.py
- [x] prepare_electricity_demand.py
- [x] prepare_feedin.py
- [x] prepare_heat_demand.py
- [x] prepare_re_potential.py
- [x] process_re_potential.py

Check if all headers are of the same form
- [x] build_datapackage.py
- [x] join_scenarios.py
- [x] optimize.py
- [x] plot_dispatch.py
- [x] plot_joined_scalars.py
[...]

PR #83, opened 2021-10-27
Features/prepare electricity demand

juliusmeier:
This PR introduces a script which downloads ENTSOE load data from OPSD and normalize it with the total energy demand of a year. Afterwards the timeseries is put in the timeseries template format.


scripts/prepare_electricity_demand.py
```diff
@@ -0,0 +1,95 @@
+import sys
+import pandas as pd
+import os
+import oemof_b3.tools.data_processing as dp
+
+
+# global variables
+OPSD_YEARS = list(
+    range(2015, 2020)
```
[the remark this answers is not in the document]
juliusmeier: That are all full years that are provided by the latest package version.

scripts/prepare_electricity_demand.py
```diff
@@ -0,0 +1,95 @@
...
+
+    """
+    # copy data frame
+    time_series = ts_raw.copy()
+
+    # extract one specific `year`
+    time_series = time_series[time_series.index.year == year]
+
+    # normalize with total electricity demand in year
+    time_series.iloc[:, 0] = time_series.iloc[:, 0] / time_series.iloc[:, 0].sum()
```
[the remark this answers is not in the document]
juliusmeier: Are you sure? Most values are displayed like 8.59e-5 for example.

scripts/prepare_electricity_demand.py
```diff
@@ -53,10 +75,10 @@ def prepare_load_profile_time_series(ts_raw, year, type, region):
     # add additional information as required by template
     ts_prepared.loc[:, "region"] = region
     ts_prepared.loc[:, "var_unit"] = TS_VAR_UNIT
-    ts_prepared.loc[:, "var_name"] = f"{type}-profile"
+    ts_prepared.loc[:, "var_name"] = "load-profile"
     ts_prepared.loc[:, "source"] = TS_SOURCE
     ts_prepared.loc[:, "comment"] = TS_COMMENT
-    ts_prepared.loc[:, "scenario"] = f"ts_{year}"
+    ts_prepared.loc[:, "scenario"] = "ts_load"
```
SabineHaas, line 81:
```suggestion
    ts_prepared.loc[:, "scenario"] = f"ts_{year}"
```

scripts/prepare_electricity_demand.py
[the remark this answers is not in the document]
SabineHaas: You are right! Thanks for the hint

scripts/prepare_electricity_demand.py
```diff
@@ -0,0 +1,117 @@
+r"""
+Inputs
+-------
+url_opsd : str
+    url of raw opsd data of electricty load as .csv
+output_file : str
+    path of output file with prepared data as .csv
```
jnnr, line 7:
```suggestion
url_opsd : str
    url of raw opsd data of electricty load as .csv
output_file : str
    ``results/_resources/load_profile_time_series.csv``: path of output file with prepared data as .csv
```

jnnr: According to #82, we want to include the filenames in the header, so that they are shown in the docs, which helps the user.

Snakefile
```diff
@@ -70,6 +73,15 @@ rule prepare_feedin:
     shell:
         "python {input.script} {input.wind_feedin} {input.pv_feedin} {output}"
 
+rule prepare_electricity_demand:
+    input:
+        opsd_url=HTTP.remote("https://data.open-power-system-data.org/time_series/2020-10-06/time_series_60min_singleindex.csv"),
+        script="scripts/prepare_electricity_demand.py"
+    output:
+        "results/_resources/load_profile_time_series.csv"
```
jnnr, line 81:
```suggestion
        "results/_resources/load_profile_electricity.csv"
```

jnnr: There will be other load profiles which have to reside in different files to have unique outputs in snakemake, therefore this name should be specific.

scripts/prepare_electricity_demand.py
```diff
@@ -0,0 +1,117 @@
...
+
+# global variables
+OPSD_YEARS = list(
+    range(2015, 2020)
+)  # opsd load time series are prepared for these years
+REGIONS = ["BB", "B"]
+TS_VAR_UNIT = "None"
+TS_SOURCE = "https://data.open-power-system-data.org/time_series/2020-10-06"
+TS_COMMENT = "DE_50hertz actual load data"
+LOAD_DATA = "DE_50hertz_load_actual_entsoe_transparency"
```
jnnr, line 37:
```suggestion
COL_SELECT = "DE_50hertz_load_actual_entsoe_transparency"
```

jnnr: The name is not really telling. Just a suggestion.

scripts/prepare_electricity_demand.py
```diff
@@ -0,0 +1,117 @@
...
+
+if __name__ == "__main__":
+    url_opsd = sys.argv[1]
+    output_file = sys.argv[2]
+
+    # initialize data frame
+    time_series_df = pd.DataFrame()
+
+    # download raw time series from OPSD
+    ts_raw = pd.read_csv(url_opsd, index_col=0)
```
jnnr, line 94: Before this download, you could check if the data does already exist in `raw`. If yes, you do not need to download. If no, you download, and if the option "SAVE_RAW_DATA" is True, save the raw data. I thought you mentioned you already had this implemented. Should be quick but may be helpful later

jnnr (commented):
Thanks for your effort, @juliusmeier and @SabineHaas! This already looks very good. Only some comments that should be quick to implement - after this we can merge.

scripts/prepare_electricity_demand.py
jnnr: Ok, now I got it. This already happens via snakemake. You have to set `keep_local=True` in the HTTPS.remote() in the snakerule. Pretty handy. Can you write a sentence in the header to document this.

I believe that it is no mistake to make keep_local=True the default, so please add that in the Snakefile.

juliusmeier:
The downloaded timeseries data from OPSD is currently just put in the oemof-B3 directory.
@jnnr do you have an idea how I can move it to the `raw` directory?

jnnr (approved):

PR #86, opened 2021-11-04
Only scalars in _resources

SabineHaas:
With this PR all intermediate results of renewable energy potential processing is moved to one directory (`RE_potential`). This leaves `_resources` with only data in the format of `scalars.csv`.


jnnr (approved):
Very good!

PR #89, opened 2021-11-09
Add installation advice for geopandas under windows in getting_starte…

### LINKED ISSUE #56: Geopandas missing in setup.py
Had to install it to run the rule `prepare_conv_pp`

MaGering:
This PR picks up a issue described in #56 by giving instructions on how to install geopandas on Windows to the `Getting started` section in the docs.


docs/getting_started.rst
```diff
@@ -46,6 +46,40 @@ If you have installation problems, consider opening an
...
+Simply type
+
+::
+
+    conda install --channel conda-forge geopandas
+
+in the Anaconda prompt
+
+
+2) Creating a new environment
```
jnnr, line 64: Is using an environment really an alternative, or should it be rather our default advice?

MaGering: I agree with you that it should be our default advice. I tested both ways. Both lead to successfull installations of geopandas on windows for me, whereas with the second way I can't understand what the commands exactly do.

I noticed that you can not use 
```
    conda install python=3 geopandas
```

without 
```
    conda config --env --add channels conda-forge

    conda config --env --set channel_priority strict
```
before.

So I would suggest to say that it is recommended for windows users to install conda in a clean environment either with

```
    conda install --channel conda-forge geopandas
```
or:
```
    conda config --env --add channels conda-forge

    conda config --env --set channel_priority strict

    conda install python=3 geopandas
```

What do you think @jnnr ?

MaGering: I simplified the explanation to the installation with `conda-forge channel` in commit https://github.com/rl-institut/oemof-B3/pull/89/commits/adb9bb5162b003a51191eb10b82a46cf2915e230.

docs/getting_started.rst
```diff
@@ -46,6 +46,19 @@ If you have installation problems, consider opening an
 `issue <https://github.com/rl-institut/oemof-B3/issues>`_.
 
 
+How to install geopandas under Windows
+--------------------------------------
+The installation of geopandas on Windows can be challenging. According to the geopandas documentation (https://geopandas.org/getting_started/install.html) there are multiple ways to install it. We recommend to use the conda-forge channel:
```
jnnr, line 51: Can you begin with a sentence explaining for what geopandas is needed?

jnnr: Currently, it is only used for preparation of conventional plant data. There could be other uses in the future. Thus: "Geopandas is necessary for a small subset of the modeling steps. Therefore it is not part of the default requirements." Or similar

jnnr (commented):
Thanks for this improvement!

docs/getting_started.rst
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/89/commits/1480f7d609c01e07c3bc9a239d73d21f39c3854b.

jnnr (approved):

Issue #66, opened 2021-09-22
New plots are missing in report


jnnr:
To report scenarios with all represented sectors, include these:
* [ ] heat dispatch
* [ ] scalar plot of capacities
* [ ] scalar plot of energy production

henhuy:
You were almost there @jnnr ...
There was a leading "." in path. after removing, pandoc replaces missing images with caption.

PR #93, opened 2021-11-24
Plot scalar results

jnnr:
Introduces a new rule "plot_scalar_results" that draw scalar result bar plots for single scenarios. You can choose to stack the bars by setting `stacked=True` in `plot_grouped_bar`. Needs https://github.com/rl-institut/oemoflex/pull/45

To test it, make sure to use the mentioned branch on oemoflex. 
* Generate some scalar plots with `snakemake -j1 results/toy-scenario/plotted/scalars`. 
* To test the integration in the report, run `snakemake -j1 results/toy-scenario/report`.

TODO 
* [x] Nicer appearance
* [x] Plot other variables (see comment below)
* [x] Check if and how it can be combined with the plot functions of joined scenarios. Done - there is only one script for both.
* Problem: Include in reports fails if file is not present. How can we deal with that? Will be addressed in #66 

Address separately
- all variable costs by region and technology
- all storage losses + output by carrier and region (This would need a new functionality: Plotting different var_names together)


jnnr:
Short brainstorm on what plot we may need:
- Plot all flows or all capacities/invest of power into electricity bus/heat bus etc
- Plot all storage losses + output by carrier and region
- Plot all variable costs by region and technology

MaGering (changes requested):
I get a `MissingInputException in line 343` with the following error message running
`snakemake -j1 results/toy-scenario/plotted/scalars`:

```
Missing input files for rule plot_scalar_results:
results/toy-scenario/postprocessed/scalars.csv
```
Also, when I run `snakemake -j1 results/toy-scenario/postprocessed`, it leads me to `AttributeError: 'NoneType' object has no attribute 'is_potentially_variable`:

```
ERROR: Constructing component 'Transformer.relation_build' from data=None
    failed: AttributeError: 'NoneType' object has no attribute
    'is_potentially_variable'
Traceback (most recent call last):
  File "scripts/optimize.py", line 50, in <module>
    m = Model(es)
...
```

jnnr:
> I get a `MissingInputException in line 343` with the following error message running `snakemake -j1 results/toy-scenario/plotted/scalars`:
> 
> ```
> Missing input files for rule plot_scalar_results:
> results/toy-scenario/postprocessed/scalars.csv
> ```

Resolved in ec05d1e. Now snakemake should trigger a model run when you try to plot results that have not been created.

> Also, when I run `snakemake -j1 results/toy-scenario/postprocessed`, it leads me to `AttributeError: 'NoneType' object has no attribute 'is_potentially_variable`:
> 
> ```
> ERROR: Constructing component 'Transformer.relation_build' from data=None
>     failed: AttributeError: 'NoneType' object has no attribute
>     'is_potentially_variable'
> Traceback (most recent call last):
>   File "scripts/optimize.py", line 50, in <module>
>     m = Model(es)
> ...
> ```

Could not reproduce this. Looks like a component of the model is missing some data?

MaGering:
Running `snakemake -j1 results/toy-scenario/report` leads me to an error because `pandoc` is missing as package. It should be added to `install_requires` in `setup.py` right?

MaGering:
After installing `pandoc` I get the following error message running `snakemake -j1 results/toy-scenario/report`:
[report_exit_status_99.txt](https://github.com/rl-institut/oemof-B3/files/7940695/report_exit_status_99.txt)

jnnr:
> Running `snakemake -j1 results/toy-scenario/report` leads me to an error because `pandoc` is missing as package. It should be added to `install_requires` in `setup.py` right?

You `pandoc` do not install pandoc via `pip`, so it should not be listed in setup.py. Need to find another way to define these requirements.

jnnr:
> After installing `pandoc` I get the following error message running `snakemake -j1 results/toy-scenario/report`: [report_exit_status_99.txt](https://github.com/rl-institut/oemof-B3/files/7940695/report_exit_status_99.txt)

Yes, the report fails if some images are missing. This is an open todo (see first post).

SabineHaas:
You could install `pandoc` from github https://github.com/jgm/pandoc using the master branch or a specific version (tag). 
For a "how-to" see oemof-B3 setup.py where oemoflex ist installed from github, as well.

jnnr:
I am not sure if that works? Other than oemoflex, pandoc is written in haskell and needs to be compiled. (https://github.com/jgm/pandoc/blob/master/INSTALL.md#getting-the-pandoc-source-code)

SabineHaas (commented):
@jnnr you have asked my feedback concerning the plots. Thank you, first of all, for improving the plots! They already look very nice. 
I have tested the plots with scenario `base-2050`, toy-scenario ran into an error.

To my mind, my comments are not so urgent that they have to be faced in this PR or for the first release.


scalars:
- I like that for the y-axis a convenient unit seems to be chosen automatically! A bit strange looks the '0 Wh` or `0 W` to my mind. Maybe there's a solution to force 0 to adapt the unit of the next higher value. 
- It would be useful for readers to add values to the stacked plots. I have added two examples below from Dena-Leitstudie 2021 and Prognos 2021.
- For the model I think B/BB and the heading are quite right, as they are related to input data and output data. However, for reports I think it would be nice to actually write the full region name and remove the heading as this will be separate in a report. My suggestion is to add a second set of plots for reports without heading and with full region names.
- For your information: I only got plots for the electricity bus. I guess plots for the other busses will be added in the future.

dispatch:
- `snakemake -j1 results/base-2050/plotted/dispatch` gives me `ValueError: need at least one array to concatenate` in `oemoflex/tools/plots.py, line 344, y = np.vstack(y)`

--------
Dena2021

Prognos2021

---------

> Address separately
> 
>     * all variable costs by region and technology
> 
>     * all storage losses + output by carrier and region (This would need a new functionality: Plotting different var_names together)
> 
>     * Check if and how it can be combined with the plot functions of joined scenarios

Should this be addressed in another PR? Then you could move these points to a new issue.

jnnr:
@SabineHaas: Thanks for your review. I opened a new issue #154 to collect all ideas that are still open.

PR #95, opened 2021-12-08
Unify resource names

jnnr:
This PR unifies the names of the resources according to the convention:

`ts/scal_<general_description>_<details>.csv`

I did not change `prepare_re_potential` because I do not know if I this applies to all files in `RE_potential`. Can you give your feedback @SabineHaas?

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -100,14 +100,6 @@ rule prepare_scalars:
     shell:
         "python {input.script} {input.raw_scalars} {output}"
 
-rule build_datapackage:
-    input:
-        "scenarios/{scenario}.yml"
-    output:
-        directory("results/{scenario}/preprocessed")
-    shell:
-        "python scripts/build_datapackage.py {input} {output}"
-
 rule prepare_re_potential:
     input:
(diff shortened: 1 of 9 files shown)


SabineHaas:
> I did not change `prepare_re_potential` because I do not know if I this applies to all files in `RE_potential`.

Yes I think it's fine like that, as files in `RE_potential` are no resources to the model but intermediate outcome.

SabineHaas (commented):
Should rule `prepare_scalars` already work? I get:
```
  File "scripts/prepare_scalars.py", line 76, in annuise_investment_cost
    assert isinstance(wacc, float)
AssertionError
```

Everything else looks good!

jnnr:
> Should rule `prepare_scalars` already work? I get:
> 
> ```
>   File "scripts/prepare_scalars.py", line 76, in annuise_investment_cost
>     assert isinstance(wacc, float)
> AssertionError
> ```

I fixed this in a commit on another branch. Cherry-picked it here to fix the problem.

jnnr:
The fix of `load_b3_scalars` occurred to be more complicated. I fixed this here even if I later realised that a separate branch would have been appropriate.

PR #99, opened 2021-12-15
Features/geopandas installation

### LINKED ISSUE #56: Geopandas missing in setup.py

MaGering:
This PR resolves #56 by adding `geopandas` to `extras_require` in `setup.py` and as module to import in `prepare_conv_pp.py`. The import is done within an exception. In case the module is not installed, an `ImportError` is raised.

### DIFF
--- a/scripts/prepare_conv_pp.py
+++ b/scripts/prepare_conv_pp.py
@@ -34,16 +34,8 @@ import sys
 import pandas as pd
 import yaml
 
-try:
-    import geopandas as gpd
-
-except ImportError:
-    print(
-        "No module named 'geopandas'. You need to install 'geopandas' in order to use this script."
-    )
-
-import oemof_b3.tools.geo as geo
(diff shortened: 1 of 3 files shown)


scripts/prepare_conv_pp.py
```diff
@@ -34,6 +34,14 @@
...
 import yaml
 
+try:
+    import geopandas as gpd
+
+except ImportError:
+    print(
+        "No module named 'geopandas'. You need to install 'geopandas' in order to use this script."
+    )
+
```
jnnr: The actual import happens in oemof_b3.tools.geo, so the try except has to happen there.

jnnr: Done in a282a0b

jnnr (approved):
Thanks for your work!

PR #101, opened 2021-12-22
Feature/vehicle charging ts

SabineHaas:
Changes in this PR

- added electric vehicle charging demand profiles with rule `prepare_vehicle_electricity_demand` and script `prepare_vehicle_electricity_demand.py`
- added test for function `prepare_vehicle_time_series`
- until now charging strategy is "greedy" as _spiceEV_ is not made for such big regions (rather for single "Fuhrparks")
- added smoothing of the profiles "home" and "work", charging strategy "balanced"

Notes
- _simBEV_ hasn't finished all years for each region, yet, but this PR works for all files that will follow with the same structure


scripts/prepare_vehicle_electricity_demand.py
```diff
@@ -0,0 +1,115 @@
...
+
+    """
+    # initialize data frame
+    df = pd.DataFrame()
+
+    # loop through files in `input_dir`
+    for filename in os.listdir(input_dir):
+        # get year and region from file name
+        split_filename = os.path.splitext(os.path.basename(filename))[0].split("_")
+        region, year = split_filename[2], int(split_filename[3])
```
jnnr, line 62: This could be a function

scripts/prepare_vehicle_electricity_demand.py
```diff
@@ -0,0 +1,115 @@
...
+        path = os.path.join(input_dir, filename)
+
+        # read data from file, copy and superfluous drop last time step
+        ts_raw = pd.read_csv(
+            path, index_col=1, sep=";", decimal=",", parse_dates=True
+        ).drop(columns=["Unnamed: 0"], axis=1)
+        ts = ts_raw.copy()
```
jnnr, line 69: As you do not use ts_raw later, this copy seems unnecessary. Correct me if I am wrong.

scripts/prepare_vehicle_electricity_demand.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # stack time series and add region
+        ts_stacked = dp.stack_timeseries(ts_df).rename(columns={"var_name": "scenario"})
+        ts_stacked.loc[:, "region"] = region
+
+        # add to `df`
+        df = pd.concat([df, ts_stacked], axis=0)
+
+    # prepare index
+    df.index = range(0, len(df))
```
jnnr, line 91:
```suggestion
    df.reset_index(inplace=True, drop=True)
    df.index.name = "id_ts"
```

jnnr (commented):
Thanks for your work. Looks good already! I could run the rule without problems. Added some little suggestions. The test fails, though, because the order of the dataframe is reversed wrt the default. We should discuss if we want to test functions defined in scripts at all first.

SabineHaas:
> Thanks for your work. Looks good already! I could run the rule without problems. Added some little suggestions. 

Thank you for your review! 

> The test fails, though, because the order of the dataframe is reversed wrt the default. We should discuss if we want to test functions defined in scripts at all first.

hmm but all checks have passed in the CI - maybe you are using another version of pandas locally

I will work on the smoothing in this PR as it is still quite small.

scripts/prepare_vehicle_electricity_demand.py
SabineHaas: Ah yes, I never know, when the copy warning is coming.. but you are right, it's not necessary.

scripts/prepare_vehicle_electricity_demand.py
SabineHaas: However, it is only used once. Or do you mean everything in the for loop could be in a separate function?

scripts/prepare_vehicle_charging_demand.py
```diff
@@ -0,0 +1,189 @@
...
+    # add additional information as required by template
+    ts_prepared.loc[:, "var_unit"] = TS_VAR_UNIT
+    ts_prepared.loc[:, "var_name"] = "vehicle-charging-profile"
+    ts_prepared.loc[:, "source"] = TS_SOURCE
+    ts_prepared.loc[:, "comment"] = TS_COMMENT
+
+    return ts_prepared
+
+
+def balance_profiles(df):
```
SabineHaas, line 130: This function is not generic but created for better overview in `prepare_vehicle_time_series`; better include it as function in function `prepare_vehicle_time_series`?

tests/_files/test_vehicle_ts/res_SimBEV_Berlin_2015.csv
```diff
@@ -0,0 +1,13 @@
+;time;sum CS power;sum UC work;sum UC business;sum UC school;sum UC shopping;sum UC private/ridesharing;sum UC leisure;sum UC home;sum UC hub
+0;01.01.2015 00:00;6011,442277;113,7397321;0;0;22;132,2632421;813,1647798;4810,274522;120
+1;01.01.2015 00:15;5834,261506;43,29294752;0;0;0;245,5309077;664,090325;4881,347325;0
```
SabineHaas, line 3: The test data could have more intuitive values, to be able to check the result manually. I haven't done this, yet, as I remember that you were not sure about keeping all tests of single functions @jnnr.

jnnr (commented):
Thanks for this features, @SabineHaas! Everything is quite clean and thecomments help, so I code read your code quite well! Tested it again and it works.

scripts/prepare_vehicle_electricity_demand.py
SabineHaas: I'm not used to doing that but it looks indeed more tidy.

scripts/prepare_vehicle_charging_demand.py
SabineHaas: I keep it outside as it is indeed quite big

SabineHaas:
Thanks @jnnr  for your constructive feedback! :+1:

scripts/prepare_vehicle_electricity_demand.py
```diff
@@ -0,0 +1,115 @@
...
+    for filename in os.listdir(input_dir):
+        # get year and region from file name
+        split_filename = os.path.splitext(os.path.basename(filename))[0].split("_")
+        region, year = split_filename[2], int(split_filename[3])
+        path = os.path.join(input_dir, filename)
+
+        # read data from file, copy and superfluous drop last time step
+        ts_raw = pd.read_csv(
+            path, index_col=1, sep=";", decimal=",", parse_dates=True
+        ).drop(columns=["Unnamed: 0"], axis=1)
```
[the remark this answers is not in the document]
jnnr: I just find the name of the column "Unnamed: 0" quite obscure and unrelated to the code. It is an artifact. Stating the columns that are actually relevant would be less arbitrary and more related to your script.

scripts/prepare_vehicle_charging_demand.py
```diff
@@ -0,0 +1,189 @@
...
+    r"""
+    Smoothes profiles "work" and "home" of `df` between specific hours (see global variables).
```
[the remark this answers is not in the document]
jnnr: Thanks for the extra explanation. Can you please add another sentence explaining how the smoothing works precisely (how is the math)?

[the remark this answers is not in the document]
jnnr: It seems that charging strategy "balanced" is a name that the user is supposed to know. To me, it seems more appropriate to explain the math and state that we name it charging strategy balanced.

jnnr (approved):
Thanks for your work, @SabineHaas! I consider this ready to be merged. If you can take the time to explain how the balanced charging strategy translates into math, this would be a benefit for the docs.

scripts/prepare_vehicle_charging_demand.py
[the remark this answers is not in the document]
SabineHaas: Good point! done in [50b1813](https://github.com/rl-institut/oemof-B3/pull/101/commits/50b181328a0a99e877c725f87a24e05cbeede6ed)

SabineHaas:
> Thanks for your work, @SabineHaas! I consider this ready to be merged. If you can take the time to explain how the balanced charging strategy translates into math, this would be a benefit for the docs.

Thank you for your feedback! I have adapted explanations in [50b1813](https://github.com/rl-institut/oemof-B3/pull/101/commits/50b181328a0a99e877c725f87a24e05cbeede6ed) and [e4a2e6a](https://github.com/rl-institut/oemof-B3/pull/101/commits/e4a2e6a4031769d288cfc4f8dd47dfbdd318a4c9).
I have also renamed the functions to names I find more exact.

jnnr:
Thanks for the extra documentation. That helps!

PR #102, opened 2021-12-22
Constraints for gas/power relation in heat busses

### LINKED ISSUE #97: Electricity/gas relation constraint
Add constraint for share between supply from electricity (WP, res. PtH) and gas (boiler, CHP x 2) for Heat central and decentral. 
The share has to be met at every time step of the simulation.

Info: https://oemof-solph.readthedocs.io/en/latest/reference/oemof.solph.html#oemof.solph.constraints.equate_variables

SabineHaas:
Fix #97, #184

Use `scalars_base_2050_PR102.csv` and copy content into `scalars_base_2050.csv` to test this PR.

To test, run `snakemake -fj1 results/base-2050/postprocessed/`

Changes in this PR:

- Added function `add_electricity_gas_relation_constraints()` which adds the constraints for gas/power relation for heat busses and all regions  
- Added preprocessing for gas/power relations (and emission_limit), saving them to "additional_scalars.csv"
- Added functions  `get_electricity_gas_relations()` and `get_bpchp_output_parameters()` to `optimize.py`
- Added function `add_output_parameters_to_bpchp()` that adds keywords for electricity/gas relation constraint to backpressure CHPs.  This is necessary as oemof.tabular does not support `output_parameters` of these components, yet.
- Added function `save_additional_scalars` to `build_datapackage`

Changes in `basic-scenario_PR116.csv`
- Added keywords 'gas_central-<region>' and 'electricity_central-<region>' to pth, ch4-boiler and backpressure CHPs
- Added `electricity_gas_relation` for heat_central and heat_decentral busses and regions B and BB.

To use the constraint copy [`equate_flows.py`](https://github.com/oemof/oemof-solph/blob/features/equate-flows/src/oemof/solph/constraints/equate_variables.py) of oemof.solph into `/tools` directory of `oemof-B3`

Possible todo or move to issue:
- :x: adding the region abbreviation to the keyword could be done automatically by `expand_regions()` in `build_datapackage.py`. For BB-B energy system this is not necessary but when using more regions this would be nice. 
- [x] After #91 is merged: Add keywords to boiler-small, heatpumps and check futher components

TODO
- [x] How to add keywords to component with two output flows?
- [x] Check functionality
- [x] `WARNING - Implicitly replacing the Component attribute equate_flows_build...`


jnnr:
I added a new constraint in oemof-solph that allows to equate flows: https://github.com/oemof/oemof-solph/pull/817

jnnr:
> @jnnr I have just realized: For using this new oemof.solph functionality we have to install the current branch (later oemof 5)

Yes, we are depending on oemof.tabular to get the release done. Before that happens, we need to use a workaround.

SabineHaas:
> * How to add keywords to component with two output flows?

@jnnr if you know how to add a keyword to one flow of a components with two output flows (CHP) I would be thankful for a hint. If not, no problem, then I'll try to find out!

scripts/optimize.py
```diff
@@ -28,29 +35,124 @@
...
+    # rename column 'name' as is collides with iterrows()
+    parameters.rename(columns={"name": "name_"}, inplace=True)
+    for i, element in parameters.iterrows():
+        # get heat bus the component is connected to
+        bus = energysystem.groups[element.name_].heat_bus
+
+        # get keyword and boolean value
+        split_str = element.var_value.split('"')
+        keyword = split_str[1]
+        value = bool(split_str[2].split("}")[0].split()[1])
```
SabineHaas, line 133: @jnnr I actually thought that `load_b3_scalars()` turned the `var_value` of output_parameters and input_parameters into a dictionaries already. But seems like that's not the case. Here it is a string.

SabineHaas:
@jnnr I still want to test this PR to check whether the results look as expected, but you can already get an idea about how I implemented the functionalities.

SabineHaas:
When adding the constraint here https://github.com/rl-institut/oemof-B3/blob/3d358bd276ac8d533f44d32e857b334e455025d6/scripts/optimize.py#L148-L153 I get the following warning:
```
WARNING: Implicitly replacing the Component attribute equate_flows
    (type=<class 'pyomo.core.base.constraint.IndexedConstraint'>) on block
    Model with a new Component (type=<class
    'pyomo.core.base.constraint.IndexedConstraint'>). This is usually
    indicative of a modelling error. To avoid this warning, use
    block.del_component() and block.add_component().
```

@jnnr did you get these in your tests in oemof.solph as well? Could you look into this, please?

jnnr:
> When adding the constraint I get the following warning:
> 
> ```
> WARNING: Implicitly replacing the Component attribute equate_flows
>     (type=<class 'pyomo.core.base.constraint.IndexedConstraint'>) on block
>     Model with a new Component (type=<class
>     'pyomo.core.base.constraint.IndexedConstraint'>). This is usually
>     indicative of a modelling error. To avoid this warning, use
>     block.del_component() and block.add_component().
> ```

This was due to the constraint being set several times with the same name. Fixed with [27edb6c](https://github.com/rl-institut/oemof-B3/pull/102/commits/27edb6c00526de95eec8e8293deb47c5fa5f6806)

scripts/optimize.py
jnnr: If this should be done in load_b3_scalars, you can open an issue.

.gitignore
```diff
@@ -13,3 +13,4 @@ build
 gurobi.log
 data.open-power-system-data.org/
 /oemof_b3/config/.secrets.yaml
+/oemof_b3/tools/equate_flows.py
```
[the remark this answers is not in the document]
SabineHaas: done in [fbd19ee](https://github.com/rl-institut/oemof-B3/pull/102/commits/fbd19ee296b7f26a38300e9093477272f78db57e)

henhuy:
Hi, reviewing so many changes is difficult for me, especially as I'm not that into oemof-B3 structure yet, sry.
Regarding the code, I have a more general question:
in _optimze.py_ you are adding outputs (line 226) and constraints (line 235) to ES/model after building the datapackage. This seems like a strange workaround for me? My spontaneous thought would be to create a new facade (i.e. inherited from Backpressure) with output logic added within the facade model and use this custom BPCHP instead of builtin BHCHP?!
Thus, users would have "clean" datapackages - or am I missing something?

jnnr:
>  in _optimze.py_ you are adding outputs (line 226) and constraints (line 235) to ES/model after building the datapackage. This seems like a strange workaround for me? My spontaneous thought would be to create a new facade (i.e. inherited from Backpressure) with output logic added within the facade model and use this custom BPCHP instead of builtin BHCHP?! Thus, users would have "clean" datapackages - or am I missing something?

Thanks for your suggestion. I agree that the way to go would be to either adapt the facade in oemof.tabular directly or, if this is unwanted, create a custom facade that inherits from backpressure.

For the constraints, we could do something similar, an extension of oemof.tabular: Facades not only for components, but also for constraints. Would require a bit more design, but would be an important feature!

jnnr (approved):
Hi @SabineHaas, thanks again for this feature. I tested it and described the results in our log. There is only one thing to check before we merge. As the scalars data structure changed while this feature was developed, the assumptions are not taken from the files as described in the first post in this PR. Can you confirm that all data has migrated into the new files (i.e. el_gas_constraint.csv, emission_constraint.csv)? If that is the case, this is ready.

SabineHaas:
> Hi @SabineHaas, thanks again for this feature. I tested it and described the results in our log. There is only one thing to check before we merge. As the scalars data structure changed while this feature was developed, the assumptions are not taken from the files as described in the first post in this PR. Can you confirm that all data has migrated into the new files (i.e. el_gas_constraint.csv, emission_constraint.csv)? If that is the case, this is ready.

Thank you for reviewing and testing @jnnr ! The el_gas_constraint.csv and emission_constraint.csv are up-to-date.

PR #106, opened 2022-01-10
Fixes/connect snakemake dag

jnnr:
This PR introduces some fixes of the DAG defined in the snakefile. You can check the DAG visually by plotting it, e.g.: `snakemake --dag plot_all_scenarios | dot -Tsvg > dag.svg`

TODO/Done
* [x] plot_all_scenarios: Obsolete (should change examples to scenarios
* [ ] plot_grouped_scenarios: DAG starts only after joining scalar results
* [ ] plot_all_resources does not work: missing input file
* [ ] is the rule clean useful as it is?

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -115,7 +115,7 @@ rule prepare_re_potential:
 
 rule process_re_potential:
     input:
-        input_dir=directory("results/_resources/RE_potential/"),
+        input_dir="results/_resources/RE_potential/",
         script="scripts/process_re_potential.py"
     output:
         scalars="results/_resources/scal_power_potential_wind_pv.csv",
--- a/Snakefile
+++ b/Snakefile
@@ -125,11 +125,15 @@ rule process_re_potential:
 
 rule build_datapackage:
     input:
-        "scenarios/{scenario}.yml"
+        scalars = "results/_resources/scal_base-scenario.csv",
+        power_potential_re="results/_resources/scal_power_potential_wind_pv.csv",
+        feedin="results/_resources/ts_feedin.csv",
+        load_electricity="results/_resources/ts_load_electricity.csv",
+        scenario="scenarios/{scenario}.yml",
     output:
         directory("results/{scenario}/preprocessed")
     shell:
-        "python scripts/build_datapackage.py {input} {output}"
+        "python scripts/build_datapackage.py {input.scenario} {output}"
 
 rule optimize:
     input:
--- a/Snakefile
+++ b/Snakefile
@@ -20,7 +20,7 @@ rule plot_grouped_scenarios:
 
 rule plot_all_scenarios:
     input:
-        expand("results/{scenario}/plotted/", scenario=examples)
+        expand("results/{scenario}/plotted/", scenario=scenarios)
 
 rule run_all_examples:
     input:


SabineHaas (commented):
Thank you @jnnr , I checked the dag and have the following comments:

- Did you want to include all rules? e.g. `prepare_conv_pp` is missing, but I did not check all the rules.
- When I run a single rule (tested with `prepare_electricity_demand`) the dag adapts as expected :+1: 
- When I create `/preprocessed` (`snakemake -fj1 results/base-scenario/preprocessed`) all rules until and incl. `build_datapackage` are shown with dashed lines. 
	However, looking into `results/_resources` the outputs of these rules (`prepare_feedin`, `prepare_electricity_demand` etc.)  do not exist. 

Note: I am using "base-scenario" for this as I did not get the toy-scenarios to run with the current files in `raw`.

jnnr:
>     * Did you want to include all rules? e.g. `prepare_conv_pp` is missing, but I did not check all the rules.

`prepare_conv_pp` is only connected to `plot_resources`.

>     * When I create `/preprocessed` (`snakemake -fj1 results/base-scenario/preprocessed`) all rules until and incl. `build_datapackage` are shown with dashed lines.
>       However, looking into `results/_resources` the outputs of these rules (`prepare_feedin`, `prepare_electricity_demand` etc.)  do not exist.

I observed something similar. When I delete some of the resources, but the preprocessed data is already generated, no new run will be triggered. Could this https://snakemake.readthedocs.io/en/stable/project_info/faq.html#snakemake-does-not-trigger-re-runs-if-i-add-additional-input-files-what-can-i-do be an explanation?

jnnr:
@SabineHaas: Is there more to do here, or can we finish this up? If you have an opinion on the rule `clean` I would be happy to hear it. I never used it so far, so I consider deleting it.

SabineHaas:
@jnnr I am using the rule `clean`. Probably it's not absolutely necessary, as we can also overwrite. But I find it useful to remove all created files sometimes.

SabineHaas:
> @SabineHaas: Is there more to do here, or can we finish this up?

I don't know whether the above mentioned issue could be solved in this PR or rather in a new PR. If you merge this PR I would recommend to open a new issue with the comment above: 



> > ```
> > * When I create `/preprocessed` (`snakemake -fj1 results/base-scenario/preprocessed`) all rules until and incl. `build_datapackage` are shown with dashed lines.
> >   However, looking into `results/_resources` the outputs of these rules (`prepare_feedin`, `prepare_electricity_demand` etc.)  do not exist.
> > ```
> 
> 
> I observed something similar. When I delete some of the resources, but the preprocessed data is already generated, no new run will be triggered. Could this https://snakemake.readthedocs.io/en/stable/project_info/faq.html#snakemake-does-not-trigger-re-runs-if-i-add-additional-input-files-what-can-i-do be an explanation?

jnnr:
> I don't know whether the above mentioned issue could be solved in this PR or rather in a new PR. If you merge this PR I would recommend to open a new issue with the comment above

Did the comment address your problem? It does not seem to be an issue, but rather a peculiarity of snakemake that we should know. We can observe this and add few sentences to the docs. Ok?

SabineHaas:
On this branch, I get the following error (also when running any scenario):
```
Building DAG of jobs...
InputFunctionException in line 144 of /home/sabine/virtualenvs/umas/oemof-B3/Snakefile:
Error:
  KeyError: 'paths_scalars'
Wildcards:
  scenario=base-scenario
Traceback:
  File "/home/sabine/virtualenvs/umas/oemof-B3/Snakefile", line 137, in get_paths_scenario_input
```
Do I miss anything? I updated the raw files

jnnr:
> Do I miss anything? I updated the raw files

Thanks for finding this. Caused by an inconsistency of a key in the scenario definitions.

SabineHaas (approved):

PR #107, opened 2022-01-11
Add missing excess/shortage components

SabineHaas:
Changes in this PR:

1. Busses `h2` and `ch4` are balanced (set to True in `bus_attrs_update.yml`), as we are working with excess (export) and shortage (import) components.
2. Add h2 shortage and h2 excess (model/, labels.yml)
3. Add h2 and ch4 shortage and h2 excess to our base scenario 
4. Change label of ch4-excess to "CH4 Export" instead of "CH4 Excess"

@jnnr 1. also solves the issue of excess components not being displayed in results. 
I am not sure why they were not displayed while the busses were unbalanced. Seems like excess component of a bus is neglected (in oemoflex?) if the bus is unbalanced. I did not check further.

Needs oemoflex branch `fix/defaults_of_excess` (https://github.com/rl-institut/oemoflex/pull/48)
and `raw/base-scenario_PR107.csv`

I gave the `marginal_cost` of `excess` the dummy value 0 in the csv. Later we add negative costs (earn money by export). Is this correct @jnnr or is this value processed an thus should be positive?

### DIFF
--- a/oemof_b3/model/component_attrs_update.yml
+++ b/oemof_b3/model/component_attrs_update.yml
@@ -60,6 +60,7 @@ heat_central-excess:
     bus: heat_central
   defaults:
     marginal_cost: 0
+    input_parameters: "{}"
 
 heat_central-shortage:
   carrier: heat_central
@@ -96,6 +97,7 @@ heat_decentral-excess:
     bus: heat_decentral
   defaults:
     marginal_cost: 0
+    input_parameters: "{}"
 
 heat_decentral-shortage:
   carrier: heat_decentral
--- a/oemof_b3/model/component_attrs_update.yml
+++ b/oemof_b3/model/component_attrs_update.yml
@@ -6,27 +6,37 @@ ch4-demand:
     bus: ch4
     profile: ch4-demand-profile
 
-ch4-shortage:
+ch4-import:
   carrier: ch4
-  tech: shortage
+  tech: import
   type: shortage
   foreign_keys:
     bus: ch4
   defaults:
     output_parameters: "{}"
 
-h2-shortage:
+ch4-export:
+  carrier: ch4
+  tech: export
+  type: excess
+  foreign_keys:
+    bus: ch4
+  defaults:
+    marginal_cost: 0
+    input_parameters: "{}"
+
+h2-import:
   carrier: h2
-  tech: shortage
+  tech: import
   type: shortage
   foreign_keys:
     bus: h2
   defaults:
     output_parameters: "{}"
 
-h2-excess:
+h2-export:
   carrier: h2
-  tech: excess
+  tech: export
   type: excess
   foreign_keys:
     bus: h2
--- a/oemof_b3/model/component_attrs_update.yml
+++ b/oemof_b3/model/component_attrs_update.yml
@@ -15,6 +15,25 @@ ch4-shortage:
   defaults:
     output_parameters: "{}"
 
+h2-shortage:
+  carrier: h2
+  tech: shortage
+  type: shortage
+  foreign_keys:
+    bus: h2
+  defaults:
+    output_parameters: "{}"
+
+h2-excess:
+  carrier: h2
+  tech: excess
+  type: excess
+  foreign_keys:
+    bus: h2
+  defaults:
+    marginal_cost: 0
+    output_parameters: "{}"
+
 electricity-electrolyzer:
   carrier: electricity
   tech: electrolyzer
--- a/oemof_b3/model/component_attrs_update.yml
+++ b/oemof_b3/model/component_attrs_update.yml
@@ -32,7 +32,7 @@ h2-excess:
     bus: h2
   defaults:
     marginal_cost: 0
-    output_parameters: "{}"
+    input_parameters: "{}"
 
 electricity-electrolyzer:
   carrier: electricity
--- a/oemof_b3/model/foreign_keys_update.yml
+++ b/oemof_b3/model/foreign_keys_update.yml
@@ -1,8 +1,9 @@
 'bus': [
     'ch4-demand',
-    'ch4-shortage',
-    'h2-shortage',
-    'h2-excess',
+    'ch4-import',
+    'ch4-export',
+    'h2-import',
+    'h2-export',
     'heat_central-shortage',
     'heat_central-excess',
     'heat_central-demand',
--- a/oemof_b3/model/foreign_keys_update.yml
+++ b/oemof_b3/model/foreign_keys_update.yml
@@ -1,6 +1,8 @@
 'bus': [
     'ch4-demand',
     'ch4-shortage',
+    'h2-shortage',
+    'h2-excess',
     'heat_central-shortage',
     'heat_central-excess',
     'heat_central-demand',
--- a/oemof_b3/config/labels.yml
+++ b/oemof_b3/config/labels.yml
@@ -18,7 +18,10 @@ hydro-ror: Hydro ROR
 solar-pv: PV
 wind-onshore: Wind on
 h2-bpchp: H2 Backpressure CHP
-ch4-excess: CH4 Excess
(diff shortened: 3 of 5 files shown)


jnnr:
>     4. Change label of ch4-excess to "CH4 Export" instead of "CH4 Excess"

What function will shortage and excess have? Just to keep the problem solvable (very high costs and zero costs respectively), or rather represent Import/Export with prices representing the market?

> @jnnr 1. also solves the issue of excess components not being displayed in results. I am not sure why they were not displayed while the busses were unbalanced. Seems like excess component of a bus is neglected (in oemoflex?) if the bus is unbalanced. I did not check further.

The results should appear in any case, but could be zero when the bus is not balanced, which renders shortage and excess obsolete.

> I gave the `marginal_cost` of `excess` the dummy value 0 in the csv. Later we add negative costs (earn money by export). Is this correct @jnnr or is this value processed an thus should be positive?

Seems correct with negative values. Seems to answer my question above as well: The components represent Import/Export.

oemof_b3/model/component_attrs_update.yml
```diff
@@ -15,6 +15,25 @@ ch4-shortage:
   defaults:
     output_parameters: "{}"
 
+h2-shortage:
+  carrier: h2
+  tech: shortage
```
jnnr, line 20: Could use the tech attribute to describe the meaning of the technology (like in 'curtailment'), e.g. import

oemof_b3/model/component_attrs_update.yml
```diff
@@ -15,6 +15,25 @@ ch4-shortage:
...
+  type: shortage
+  foreign_keys:
+    bus: h2
+  defaults:
+    output_parameters: "{}"
+
+h2-excess:
+  carrier: h2
+  tech: excess
```
jnnr, line 29: Same here, could name tech 'export'

oemof_b3/model/foreign_keys_update.yml
```diff
@@ -1,6 +1,8 @@
 'bus': [
     'ch4-demand',
     'ch4-shortage',
```
jnnr, line 3: Is ch4-excess missing?

jnnr (commented):
Thanks for this! I got only some comments and questions.

oemof_b3/model/component_attrs_update.yml
SabineHaas: Good idea! 
Seems like, it's not enough to just adapt `tech`. 
However, if I also adapt the name of the component e.g. `h2-import` I get an error that a facade `import.csv` is missing. `curtailment` does not have a facade in `oemoflex`, still seems to work fine. 
Could you give me a hint on that @jnnr ?

oemof_b3/model/foreign_keys_update.yml
SabineHaas: It is already added in  `foreign_keys` of `oemoflex`

SabineHaas:
> > ```
> > 4. Change label of ch4-excess to "CH4 Export" instead of "CH4 Excess"
> > ```
> 
> What function will shortage and excess have? Just to keep the problem solvable (very high costs and zero costs respectively), or rather represent Import/Export with prices representing the market?

They will represent import and export. So, should we add another Shortage and Excess with very high and zero costs? Or can we solve this by using unbalanced busses? 
If we use Shortage and Excess they are displayed in the results (we can check e.g. by how much the system is short).

> 
> > @jnnr 1. also solves the issue of excess components not being displayed in results. I am not sure why they were not displayed while the busses were unbalanced. Seems like excess component of a bus is neglected (in oemoflex?) if the bus is unbalanced. I did not check further.
> 
> The results should appear in any case, but could be zero when the bus is not balanced, which renders shortage and excess obsolete.

That's what I would expect, as well. however, h2 and ch4 excess and shortage did not appear in the results.

> 
> > I gave the `marginal_cost` of `excess` the dummy value 0 in the csv. Later we add negative costs (earn money by export). Is this correct @jnnr or is this value processed an thus should be positive?
> 
> Seems correct with negative values. Seems to answer my question above as well: The components represent Import/Export.

Yes.

SabineHaas:
> > > What function will shortage and excess have? Just to keep the problem solvable (very high costs and zero costs respectively), or rather represent Import/Export with prices representing the market?
> > 
> > 
> > They will represent import and export. So, should we add another Shortage and Excess with very high and zero costs? Or can we solve this by using unbalanced busses? If we use Shortage and Excess they are displayed in the results (we can check e.g. by how much the system is short).
> 
> Ok. We do not need Shortage and excess in that case because the problem will always be solvable.

ah yes, export and import is unlimited of course.

oemof_b3/model/component_attrs_update.yml
SabineHaas: Ah so maybe I accidentally had changed type instead - or additionally. 
Works fine now with 3d8c49a

SabineHaas:
> > I gave the `marginal_cost` of `excess` the dummy value 0 in the csv. Later we add negative costs (earn money by export). Is this correct @jnnr or is this value processed an thus should be positive?
> 
> Seems correct with negative values. Seems to answer my question above as well: The components represent Import/Export.

With negative value (I used -1) the problem is not solved, only with positive value (I used 1). But shouldn't the problem always be solvable, no matter how much h2/ch4 export earns or costs (these are the only values I have changed)?

jnnr:
Your problem seems to be unbounded. That can happen if there is a way for the objective to diverge to negative, "endless revenues" so to speak. You should check your cost assumptions.

SabineHaas:
I had forgotten that we haven't added `capacity_potential`s to renewables, yet. I solved this by lowering the export price now, but we should add capacity potentials soon. 

I also updated `base-scenario_PR170.csv` on the server. After merging this PR I will add the changes to `base-scenario.csv`.

jnnr (approved):
Looks good, thanks @SabineHaas! I tested it - had to set the export price to zero, otherwise the problem was unbounded.

SabineHaas:
> Looks good, thanks @SabineHaas! I tested it - had to set the export price to zero, otherwise the problem was unbounded.

Did you use the up-to-date file from the server? The problem was solved with -0.1 or -0.01 for me.  (just being curious)

jnnr:
> Did you use the up-to-date file from the server? The problem was solved with -0.1 or -0.01 for me. (just being curious)

I did use the file that was on the server. I wonder why this would make a difference. Still, the model will be capable of installing infinite hydrogen gasturbine capacity and make infinity profits.

SabineHaas:
@jnnr I have saved the current `basic-scenario.csv` in the archive (`2022-01-15_base-scenario_before_adding_components_PR107.csv`) and will add the new line from this PR to `basic-scenario.csv` now.

SabineHaas:
> > Did you use the up-to-date file from the server? The problem was solved with -0.1 or -0.01 for me. (just being curious)
> 
> I did use the file that was on the server. I wonder why this would make a difference. Still, the model will be capable of installing infinite hydrogen gasturbine capacity and make infinity profits.

No, I haven't added a price for `electricity-curtailment`.
The unbound problem would happen, if infinite pv/wind/hydropower and electrolyser is installed, selling ifinite h2. 

This leads me to the question: Do we need `electricity-curtailment`, shouldn't it be `electricity-export`? Or shall we keep `electricity-curtailment` in case the export price is 0 or negative?

jnnr:
> No, I haven't added a price for `electricity-curtailment`. The unbound problem would happen, if infinite pv/wind/hydropower and electrolyser is installed, selling ifinite h2.

You are right, I mixed this up.

> This leads me to the question: Do we need `electricity-curtailment`, shouldn't it be `electricity-export`? Or shall we keep `electricity-curtailment` in case the export price is 0 or negative?

Good point. If we allow electricity exports, curtailment would only happen if the line capacites are exhausted. Let's discuss this.

PR #110, opened 2022-01-14
Fix/delimiter inconsistency

### LINKED ISSUE #75: Inconsistent delimiters in data
All delimiters of data in the different processing steps (apart from raw data) should be the same, semicolon or comma.

MaGering:
This PR resolves #75 

With this PR semicolon is used instead of comma as delimiter in scalars and time series of `oemof-B3`. [This applies to all data before optimization (comment @jnnr)]. For this purpose adaptations were made in `data_processing.py`'s:

- `load_b3_scalars`
- `load_b3_timeseries`
- `test_save_df_sc`

Furthermore tests were adapted.


MaGering:
I could not find the usage of `pd.read_csv` in 

- process_re_potential.py ~ l. 77
- prepare_feedin.py ~ l. 65

as mentioned in [here](https://github.com/rl-institut/oemof-B3/issues/75#issuecomment-945843258). It was an old state in which `pd.read_csv` has already been replaced.

However I'm not sure about these code lines in:
- [`join_scenarios.py`](https://github.com/rl-institut/oemof-B3/blob/bb8b5ad7ac424ff267d2aaf67bee855cbcbc8d25/scripts/join_scenarios.py#L28)
- [`plot_dispatch.py`](https://github.com/rl-institut/oemof-B3/blob/bb8b5ad7ac424ff267d2aaf67bee855cbcbc8d25/scripts/plot_dispatch.py#L64)
- [`plot_joined_scalars.py`](https://github.com/rl-institut/oemof-B3/blob/bb8b5ad7ac424ff267d2aaf67bee855cbcbc8d25/scripts/plot_joined_scalars.py#L141)

jnnr (commented):
Thanks for that fix, @MaGering! Looks good. The three lines you were unsure about are related to postprocessing. As the postprocessed data comes in a format defined by oemoflex, this is beyond the scope here. We may consider bringing the postprocessed data into oemof_b3-format, but not on this branch.

Can you please share the updated raw data with us so that we can test if the pipeline runs irrespective of the new delimiters?

MaGering:
I found that in some scripts `to_csv` is still used in order to save `scalars` and `timeseries` as `csv` files. I changed this in `prepare_scalars.py` and `prepare_conv_pp.py` with commit https://github.com/rl-institut/oemof-B3/pull/110/commits/32b51aefc706ea51c0a7fdb551e3add8042dc7a8. However I'm not sure about these cases in:

- [postprocess.py](https://github.com/rl-institut/oemof-B3/blob/9a8b76a48bf095c88cf157c9fa6258cf2e0ed199/scripts/postprocess.py#L43)
- [build_datapackage.py](https://github.com/rl-institut/oemof-B3/blob/9a8b76a48bf095c88cf157c9fa6258cf2e0ed199/scripts/build_datapackage.py#L264)
- [join_scenarios.py](https://github.com/rl-institut/oemof-B3/blob/9a8b76a48bf095c88cf157c9fa6258cf2e0ed199/scripts/join_scenarios.py#L49)
- [create_empty_scalars.py](https://github.com/rl-institut/oemof-B3/blob/9a8b76a48bf095c88cf157c9fa6258cf2e0ed199/scripts/create_empty_scalars.py#L138)

The first two (`postprocess.py` and `build_datapackage.py`) use `to_csv_dir` function of `oemoflex`and would therefore be associated with a modification of this external repository. How should I deal with the above cases @jnnr?

jnnr:
>     * [create_empty_scalars.py](https://github.com/rl-institut/oemof-B3/blob/9a8b76a48bf095c88cf157c9fa6258cf2e0ed199/scripts/create_empty_scalars.py#L138)

Here, you can use  `save_df`. The others depend on oemoflex's data conventions, as you describe it. Oemoflex does not allow to define the delimiter, yet. Can you please open an issue on oemoflex for that?

scripts/prepare_conv_pp.py
```diff
@@ -160,4 +160,4 @@ def set_unit(df, unit_dict):
     conv_scalars_prepared.replace(carrier_dict, inplace=True)
 
     # export prepared conventional power plant data
-    conv_scalars_prepared.to_csv(out_path, index=False)
+    conv_scalars_prepared.to_csv(out_path, sep=";", index=False)
```
jnnr, line 163: Better use `save_df` for all these instances where you save some data in oemof-b3 resources format.

jnnr: I cannot recall why it seemed necessary to ignore the index when saving. Could be obsolete.

MaGering: I wanted to, but that does not work in here because of `index=False`. `save_df` uses `index=True` as default and does not have an input parameter `index` so far, where I could change this.

MaGering:
> Here, you can use `save_df`. The others depend on oemoflex's data conventions, as you describe it. Oemoflex does not allow to define the delimiter, yet. Can you please open an issue on oemoflex for that?

I created the issue: https://github.com/rl-institut/oemoflex/issues/49

scripts/prepare_conv_pp.py
MaGering: You're right. It's obsolete. I tested this with `dp.save_df` in `prepare_conv_pp.py`, `snakemake -j1 prepare_conv_pp` and `snakemake -j1 results/_resources/plots/scal_conv_pp_var_capacity_net_el.png`.

MaGering: Adjusted with commit https://github.com/rl-institut/oemof-B3/pull/110/commits/c3b8f130069303ed5a04e62a30178a0de857f48d.

jnnr:
Tested and succeeded to recreate base-scenario.

PR #112, opened 2022-01-20
Features/plot resources timeseries

jnnr:


jnnr:
This would benefit from more clean up of plots.

PR #113, opened 2022-01-20
Fixes: Input data overview table

jnnr:
* Fixes the rounding of ints.
* Includes units
* Allows to group variables (e.g. capacities and storage capacities)

### DIFF
--- a/scripts/create_input_data_overview.py
+++ b/scripts/create_input_data_overview.py
@@ -54,6 +54,8 @@ ROUND = {
 DISPLAY_UNITS = {
     "Capacity cost": ["capacity_cost_overnight", "storage_capacity_cost_overnight"],
     "Fix OM cost": ["fixom_cost", "storage_fixom_cost"],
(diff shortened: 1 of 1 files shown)


jnnr:
@SabineHaas: I adapted the table as described above and included the new version in our latex document. Feedback is welcome, but if all seems fine I would just merge this.

SabineHaas (approved):
I have checked the table as csv and it looks good to me, thank you!

Only one more small thing: We don't do any unit transformations in oemof-B3 right? So if the unit is the same for all values of one parameter, for Latex I would rather add it wiht `[<unit>]` to the columns.

jnnr:
> I have checked the table as csv and it looks good to me, thank you!

Thanks for checking!

> Only one more small thing: We don't do any unit transformations in oemof-B3 right? So if the unit is the same for all values of one parameter, for Latex I would rather add it wiht `[<unit>]` to the columns.

Thanks. No, we don't. I agree with you, but suggest that we do this manually.

PR #121, opened 2022-02-08
Fix components in toy-scenario YAML files

MaGering:
This PR is a fix of labels which are missing and are not correct among `components` in these files:

- toy-scenario.yml
- toy-scenario-2.yml

They were forgotten merging the PR https://github.com/rl-institut/oemof-B3/pull/107.

### DIFF
--- a/scenarios/toy-scenario-2.yml
+++ b/scenarios/toy-scenario-2.yml
@@ -25,8 +25,8 @@ components:
   - ch4-boiler
   - ch4-bpchp
   - ch4-demand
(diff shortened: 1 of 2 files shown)


jnnr (approved):
Looks good! Have you tested it? If that works, this should be ready to be merged.

MaGering:
> Looks good! Have you tested it? If that works, this should be ready to be merged.

I've tested it running: 
```
snakemake -j1 results/toy-scenario/plotted
```

Issue #111, opened 2022-01-18
Timeindex set at different places


jnnr:
A timeindex is defined at different places, at least here:

- create_empty_scalars
- prepare_feedin
- build_datapackage
- plot_dispatch

Improvement: Define timeindex at one place. To plot timeframes of the results in `plot_dispatch`, use the year of the data that is plotted.

PR #122, opened 2022-02-08
Feature/use consistent timeindex

MaGering:
This PR resolves issue #111

In this PR hardcoded datetimeindex is replaced by dataspecific one.

For the two files:
- `create_empty_scalars.py` and
- `build_datapackage.py`
a datetimeindex is used from the YAML-file `base-scenario.yml`.

In
- `plot_dispatch.py`
the datetimeindex is collected from the plotted data.

Procedure in line with the above is still needed for:
- `prepare_feedin.py`

### DIFF
--- a/scenarios/base-scenario.yml
+++ b/scenarios/base-scenario.yml
@@ -5,6 +5,14 @@
 
 name: base-scenario
 
+datetimeindex:
+  start:
+    "2019-01-01"
+  freq:
+    "H"
+  periods:
+    8760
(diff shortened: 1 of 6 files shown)


MaGering:
@SabineHaas could you please help to find a way to use non hardcoded datetimeindex in `prepare_feedin.py`?

jnnr (approved):
Thanks, @MaGering, for unifying this!

Issue #115, opened 2022-01-25
Modelling of electricity export


SabineHaas:
> This leads me to the question: Do we need `electricity-curtailment`, shouldn't it be `electricity-export`? Or shall we keep `electricity-curtailment` in case the export price is 0 or negative?

Good point. If we allow electricity exports, curtailment would only happen if the line capacites are exhausted. Let's discuss this.

_Originally posted by @jnnr in https://github.com/rl-institut/oemof-B3/issues/107#issuecomment-1021041679_

SabineHaas:
- BB export should be limited due to transmission limits
- B should not have `export`

PR #123, opened 2022-02-08
Add EL export/import and h2/ch4 shortages

SabineHaas:
Address #115 
In this PR I additionally worked on H2 and CH4 components. 

This PR uses `base-scenario_PR123.csv`

Changes in this PR:
- Added H2 and CH4 shortage. (They were turned into import before. However, as we have an emission limit and the import does have an emission_factor I added a shortage to these busses additionally.) 
- Added electricity-export and import (for BB). Kept electricity shortage (B could be short).

Changes in csv (`base-scenario_PR123.csv` compared to `base-scenario.csv`):
- Added above mentioned components 
    - `BB-electricity-export` and `BB-electricity-import` including `capacity`, `input_parameters`(`emission_factor`) and `marginal_costs`  
    - `marginal_cost` of H2 and CH4 `shortage`)
- Removed emission factor from electricity shortage. Now default (`{}`) is used.

open Questions:
1. Do we want to keep H2 and CH4 export/import in B? Can Berlin export to outside of Brandenburg?
2. Any more parameters needed for electricity import and export?


SabineHaas:
@jnnr unfortunately I get an error with the new csv that I could not solve, yet.
[error.txt](https://github.com/rl-institut/oemof-B3/files/8031758/error.txt)
 It has something to do with reindexing (`reindex_like()` in `generic.py`. Do you have any idea how to solve this @jnnr ?

oemof_b3/model/component_attrs_update.yml
```diff
@@ -54,6 +72,26 @@ electricity-electrolyzer:
...
     output_parameters: "{}"
 
+electricity-export:
+  carrier: electricity
+  tech: export
+  type: excess
+  foreign_keys:
+    bus: electricity
+  defaults:
+    marginal_cost: 0
```
jnnr, line 82: I suggest that we do not set the marginal cost as a default here. IMO it will be nonzero and defined in the scalar data.

oemof_b3/model/component_attrs_update.yml
```diff
@@ -54,6 +72,26 @@ electricity-electrolyzer:
...
+    input_parameters: "{}"
+
+electricity-import:
+  carrier: electricity
+  tech: import
+  type: shortage
+  foreign_keys:
+    bus: electricity
+  defaults:
+    marginal_cost: 0
```
jnnr, line 92: Same here.

jnnr (commented):
Thanks for thinking about this. Looks good! Regarding your questions:

1. Currently, we do not have h2/ch4 transmission between BB and B, so we need import for both. It seems ok to me if we keep it at that level of detail. If we decide to model gas transmission later, we can still set the import capacity of B to zero.
2. No, as far as I can see.

jnnr:
> @jnnr unfortunately I get an error with the new csv that I could not solve, yet.

Thanks for pointing to this. Admittedly, the error message could be much more specific and helpful. It happens because there are duplicate indices in the scalar data. With a184a0a, this error is caught earlier.

oemof_b3/model/component_attrs_update.yml
SabineHaas: :+1:  I have removed it from all import/export components.

PR #132, opened 2022-02-16
Activate logging in Snakefile

### LINKED ISSUE #90: No logging yet
oemof-b3's steps should be logged.

Could use snakemakes `log` directive. To capture the STDERR and STDOUT of the scripts, a redirection is necessary according to https://snakemake.readthedocs.io/en/stable/tutorial/advanced.html#step-5-logging.

henhuy:
Closes #90


jnnr (commented):
Looks good, thanks for implementing this! Can you please give recommendations on how to get more logging messages? Does it make sense to define different logging levels (info, debug)? If you can provide an example on how to use it, his PR could be finished and the adaptions of the scripts can happen separately.

jnnr:
@henhuy: Is it possible to keep displaying stdout when running the scripts but also log it at the same time?

henhuy:
> Looks good, thanks for implementing this! Can you please give recommendations on how to get more logging messages? Does it make sense to define different logging levels (info, debug)? If you can provide an example on how to use it, his PR could be finished and the adaptions of the scripts can happen separately.

Will add some logging to the scripts...

> @henhuy: Is it possible to keep displaying stdout when running the scripts but also log it at the same time?

I will check!

henhuy:
Logging is now shown on stdout and stored in file.
Added example logging in scripts `build_datapackage`, `optimize` and `postprocess`.
Two things have to be done in order to log:
1. add `params.logfile` to snake rule and add logfile to python command
2. add logger within `__main__` part of script:
```
    logfile = sys.argv[3]  # Index is related to snake call!
    logger = config.add_snake_logger(logfile, "build_datapackage")  # name is used as logging name in file handler!
```

jnnr (approved):
Thanks, @henhuy! As I understood you, we will implement the logging in the other scripts on separate branches?

henhuy:
> Thanks, @henhuy! As I understood you, we will implement the logging in the other scripts on separate branches?

Yes, this is what I would suggest. So that we can merge basic logging support via this PR.

Moved config.py for logging to oemof_b3.config module, so that future configs (from dynaconf) are all loaded from same location...

PR #135, opened 2022-02-16
Moved pinned sphinx-rtd-theme dep to docs requirements

### LINKED ISSUE #134: Find sustainable solution for the requirements
After the merge of https://github.com/rl-institut/oemof-B3/pull/118, which according to the description was a "(dirty) fix", a sustainable solution is needed. The need arises, for example, because issue https://github.com/rl-institut/oemof-B3/issues/61 reappears after it has been fixed with PR https://github.com/rl-institut/oemof-B3/pull/96.

Background: The rendering of the points works with `sphinx-rtd-theme>=0.5.1` if this package is in the `install_requires` but not if it is listed in the `extras`.

To avoid this and other potential errors of this kind, a sustainable solution is required in which the package versions do not clash.

henhuy:
Fixes #134 

@MaGering could you check if this works? (I dont have access to oemof-b3 rtd build page)

### DIFF
--- a/docs/docs_requirements.txt
+++ b/docs/docs_requirements.txt
@@ -1,4 +1,4 @@
 -e.
 sphinx
-sphinx_rtd_theme
+sphinx-rtd-theme>=0.5.1
 sphinxcontrib_bibtex
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ def read(fname):
 
 
 extras = {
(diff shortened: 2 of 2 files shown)


MaGering:
Cool, this works! :)
https://oemof-b3.readthedocs.io/en/fix-rtd_theme/examples.html

jnnr (approved):

PR #138, opened 2022-02-16
Added lua_filter to skip missing images

### LINKED ISSUE #66: New plots are missing in report

henhuy:
Fixes #66

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -200,7 +200,8 @@ rule report:
         shell(
         """
         pandoc -V geometry:a4paper,margin=2.5cm \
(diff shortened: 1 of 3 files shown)


jnnr:
Tested again with `base-scenario`. Same problem - the filter removes all files, even those that are there.

jnnr (approved):
Tested with pandoc 2.17 - it works. Thanks, @henhuy!

PR #141, opened 2022-02-18
Feature/hydro ts all years

### LINKED ISSUE #133: Hydropower profiles for all years
Hydropower profile exists only for one year, so far.

- [x] Save profile for each year (like weather data, pv, wind..)
- [x] stretch time series for leap year

SabineHaas:
Fix #133

changes in this PR:
- Save profile for each year (like weather data, pv, wind..)
- Stretch time series for leap years

### DIFF
--- a/scripts/prepare_feedin.py
+++ b/scripts/prepare_feedin.py
@@ -33,10 +33,9 @@ import os
 import oemof_b3.tools.data_processing as dp
 
 # global variables
+YEARS = list(range(2010, 2020))
 # specific to wind and pv time series
-RE_NINJA_YEARS = list(
-    range(2010, 2020)
-)  # re ninja wind+pv time series are prepared for these years
+
 NUTS_DE30 = "DE30"
 NUTS_DE40 = "DE40"
 RENAME_NUTS = {NUTS_DE30: "B", NUTS_DE40: "BB"}
@@ -47,7 +46,6 @@ TS_COMMENT = "navigate to country Germany"
 REGIONS = ["BB", "B"]
 TS_SOURCE_ROR = "https://zenodo.org/record/1044463"
 TS_COMMENT_ROR = "Isolated ror availability time series from DIW data"
-YEAR_ROR = 2017
 
 
 def prepare_wind_and_pv_time_series(filename_ts, year, type):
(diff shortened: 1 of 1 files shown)


jnnr (approved):
Thanks, @SabineHaas! The code looks fine and it runs on my system.

SabineHaas:
> Thanks, @SabineHaas! The code looks fine and it runs on my system. Flake8 complains at one line - with that fixed you can merge.

Thank you @jnnr 

Concerning the flake8 complaint: I tried `'if leap_year is True:'` or `'if leap_year:'` as the error message suggests, but then, although `leap_year` is True it just passes the if statement without entering. 
Do you have any idea why this could be?

SabineHaas:
@henhuy  although this is already merged.. do you know how to use `if cond` here: https://github.com/rl-institut/oemof-B3/blob/3a090e09cf738ca5d4d3201ab25083f9e4e1c792/scripts/prepare_feedin.py#L137

If I use `if leap_year:`, although `leap_year` is True it just passes the if statement without entering.

henhuy:
Sounds strange?! Can you tell me what snake command to run in order to call the function with `leap_year=true` ?

SabineHaas:
Yes, you can run `snakemake -fj1 prepare_feedin`

henhuy:
On my side, everything works fine with `if leap_year`.
Here output of `print(leap_year, type(leap_year))` and whether condition was interpreted as true or false:
grafik

henhuy:
The only thing I can think of, why this could behave differently on your PC is that leap_year is of type `numpy.bool_` instead of real `bool` - and maybe this is treated different depending on numpy version??? But normally this should work.

jnnr:
> The only thing I can think of, why this could behave differently on your PC is that leap_year is of type `numpy.bool_` instead of real `bool` - and maybe this is treated different depending on numpy version??? But normally this should work.

This rings a bell - we had a simlar or the same problem in oemof-flexmex and used the #noqa directive here https://github.com/modex-flexmex/oemof-flexmex/blob/e6fa5d542dac45315e8cdf70bb6192717e36c2cf/oemof_flexmex/postprocessing.py#L876

PR #153, opened 2022-02-24
Allow "additional_scalars.csv" not to exist to fix examples

### LINKED ISSUE #148: 'results/base/preprocessed/additional_scalars.csv' not created in examples
Running `snakemake -j1 run_all_examples` leads to the following error:
`FileNotFoundError: [Errno 2] File b'results/base/preprocessed/additional_scalars.csv' does not exist: b'results/base/preprocessed/additional_scalars.csv'`
This is apparently because build_datapackage.py is not executed when running the examples.

SabineHaas:
Fix #148

### DIFF
--- a/scripts/optimize.py
+++ b/scripts/optimize.py
@@ -92,7 +92,7 @@ if __name__ == "__main__":
         m = Model(es)
 
         # Add an emission constraint
-        if additional_scalars is not None:
+        if additional_scalars is not None:  # only then `emission_limit` exists
             if emission_limit is not None:
                 constraints.emission_limit(m, limit=emission_limit)
 
--- a/scripts/optimize.py
+++ b/scripts/optimize.py
@@ -73,7 +73,8 @@ if __name__ == "__main__":
 
     solver = "cbc"
 
-    # get additional scalars
+    # get additional scalars, set to None at first
+    emission_limit = None
     additional_scalars = get_additional_scalars()
     if additional_scalars is not None:
         emission_limit = get_emission_limit(additional_scalars)
@@ -92,9 +93,8 @@ if __name__ == "__main__":
         m = Model(es)
 
         # Add an emission constraint
-        if additional_scalars is not None:  # only then `emission_limit` exists
-            if emission_limit is not None:
-                constraints.emission_limit(m, limit=emission_limit)
+        if emission_limit is not None:
+            constraints.emission_limit(m, limit=emission_limit)
 
         # select solver 'gurobi', 'cplex', 'glpk' etc
         m.solve(solver=solver)
--- a/scripts/optimize.py
+++ b/scripts/optimize.py
@@ -33,10 +33,8 @@ from oemof_b3.tools import data_processing as dp
 from oemof_b3.config import config
 
 
-def get_emission_limit():
+def get_emission_limit(scalars):
     """Reads emission limit from csv file in `preprocessed`."""
-    path = os.path.join(preprocessed, "additional_scalars.csv")
-    scalars = dp.load_b3_scalars(path)
     emission_df = scalars.loc[scalars["carrier"] == "emission"].set_index("var_name")
 
     # drop row if `var_value` is None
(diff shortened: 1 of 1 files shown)


scripts/optimize.py
```diff
@@ -53,6 +51,17 @@ def get_emission_limit():
         return limit
 
 
+def get_additional_scalars():
+    """Returns additional scalars as pd.DataFrame or None"""
```
SabineHaas, line 55: This function will return more values in PR #102

scripts/optimize.py
```diff
@@ -53,6 +51,17 @@ def get_emission_limit():
...
+    filename_add_scalars = os.path.join(preprocessed, "additional_scalars.csv")
+    if os.path.exists(filename_add_scalars):
+        scalars = dp.load_b3_scalars(filename_add_scalars)
+        emission_limit = get_emission_limit(scalars)
+    else:
+        emission_limit = None
+    return emission_limit
```
jnnr, line 62: I suggest to decouple `get_additional_scalars` and `get_emission_limit`. 

In the main part of the script, let `get_additional_scalars` load the scalars and nothing more. If it returns something, call  `get_emission_limit`, passing the additional scalars. If not, do not process the emission limit.

jnnr (commented):
Thanks, @SabineHaas, for this fix. I have a suggestion that would make your code more transparent, in my view.

scripts/optimize.py
SabineHaas: done in [e8ca7ab](https://github.com/rl-institut/oemof-B3/pull/153/commits/e8ca7ab67a4ddf9a6b339d6b422817f0ac1b1689)

this leads to a warning in PyCharm that `emission_limit` can be undefined. Which I ensure not to be, as I first ask whether `get_additional_scalars` returned something and if it did `emission_limit` exists. But I am not sure whether this is good coding practice.

scripts/optimize.py
```diff
@@ -81,8 +92,9 @@ def get_emission_limit():
         m = Model(es)
 
         # Add an emission constraint
-        if emission_limit is not None:
-            constraints.emission_limit(m, limit=emission_limit)
+        if scalars is not None:
```
jnnr, line 95: You can omit checking this because you have checked it above already.

scripts/optimize.py
```diff
@@ -65,7 +73,10 @@ def get_emission_limit():
 
     solver = "cbc"
 
-    emission_limit = get_emission_limit()
+    # get additional scalars
+    scalars = get_additional_scalars()
+    if scalars is not None:
```
jnnr, line 78:
```suggestion
    if additional_scalars is not None:
```

jnnr (approved):
Thanks, @SabineHaas, this is much easier to read. Two more suggestions - apart from that this is ready.

scripts/optimize.py
jnnr: Would rename for better readability.

scripts/optimize.py
SabineHaas: I need to check it here again, as `emission_limit` does not exist if `scalars` is None. See here: https://github.com/rl-institut/oemof-B3/blob/e8ca7ab67a4ddf9a6b339d6b422817f0ac1b1689/scripts/optimize.py#L78-L79

Another option would be to add `else` to L78-L79 
```
    if additional_scalars is not None:
        emission_limit = get_emission_limit(additional_scalars)
    else:
        emission_limit = None

```

This will get quite long, as other variables will have to be set here, too, see #97.
This is why I've moved that into a function in the first place.

I would just keep the `if scalars is not None:` here a second time to save lines and for easier understanding.

jnnr: Proposal: Set `emission_limit = None` at the very begging.

scripts/optimize.py
```diff
@@ -65,7 +73,11 @@ def get_emission_limit():
 
     solver = "cbc"
 
-    emission_limit = get_emission_limit()
+    # get additional scalars, set to None at first
+    emission_limit = None
```
SabineHaas, line 77: for your information: `el_gas_relations` and `bpchp_out` will be added here in #102

Issue #109, opened 2022-01-13
Examples' naming out of date


jnnr:
Some of the values in the examples are not consistent with the rest. This leads to errors when plotting (e.g. `snakemake -j1 resutls/base/plotted`) Should rename to keep it work.

* BE -> B
* liion-battery -> liion_battery

jnnr:
The problem is rather that the script in #93, which defines the regions as constants instead of using the regions of the data.

The problem with `liion-battery` could not be reproduced.

jnnr:
There is a problem with the names of the regions when building the report (the expected regions are hard-coded). 

TODO:
* [ ] Adapt the name of the region BE to B in the examples.
* [ ] Adapt the name of the plots BE to B in the report.

monika-o:
@jnnr The above command results in the error `[Errno 2] No such file or directory: 'scenarios/base.yml'`. Could it be that the file name in the Snakefile at line 143 is not correct? I am working on the branch fix/BE-to-B which I just created from the dev-branch, so it should be identical to the dev-branch.

jnnr:
Removed this from the release milestone as it is not ready. Dropped the target rule `report_all_examples` from Snakefile in [2bba3a1](https://github.com/rl-institut/oemof-B3/pull/169/commits/2bba3a1124e4becaaa694827ede6bd908b9dee55) for now, as the plots do not appear in the report.

PR #161, opened 2022-03-02
Rename BE to B in examples

monika-o:
This PR intends to solve #109 by renaming BE into B in the examples.


monika-o:
Unfortunately, the changes so far affect the results (the postprocessed data), which I don't understand. Does anybody have an idea why this could have happened?

monika-o:
Apparently, oemof-B3 has problems with a single B as region name. I have created a new branch from dev and renamed the region once into "Berlin" and once into "B" and the results only change in the latter case.

PR #165, opened 2022-03-02
Add pre-commit to dev dependencies

SabineHaas:
@henhuy I pinned it to one version as I don't know with which schema you pinned the other version. Feel free to allow more versions of pre-commit.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -169,6 +169,14 @@ python-versions = "*"
 pycparser = "*"
 
 [[package]]
+name = "cfgv"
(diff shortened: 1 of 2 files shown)


henhuy:
In poetry you add dependencies via:
```
poetry add <package>
poetry add --dev <package>  # for dev-dependencies
```
Poetry will try to install newest package which has no conflicts with other deps in repo.

jnnr (approved):
Did not test it, but looks fine to me.

Issue #159, opened 2022-03-01
Do not use specific scalars as input data


SabineHaas:
During #158 I realized that there are rules that still use a specific scalars input file:

rule prepare_re_potential: assumptions="raw/scalars.csv",
prepare_heat_demand: scalars="raw/scalars_base.csv",

For heat demand this might be on purpose (@MaGering ?), for the re potential it should be the scenario specific scalars.

SabineHaas:
This has been solved with the new data structure

PR #171, opened 2022-03-03
Use scalars_base.csv as input for prepare_re_potential.py

SabineHaas:
address #159

### DIFF
--- a/scripts/prepare_re_potential.py
+++ b/scripts/prepare_re_potential.py
@@ -15,7 +15,7 @@ filename_kreise : str
     ``raw/lookup_table_brandenburg_kreise.csv``: Path incl. file name of lookup table Kreise and
     NUTS of Brandenburg
 filename_assumptions : str
-    ``raw/scalars_base.csv``: Path incl. file name of assumptions
+    ``raw/scalars_base_2050.csv``: Path incl. file name of assumptions
 output_dir : str
     ``results/_resources/RE_potential/``: Path of output directory where power and area potential of
     single areas and of "Landkreise" is saved


Snakefile
```diff
@@ -143,7 +143,7 @@ rule prepare_re_potential:
         pv_road_railway="raw/area_potential/2021-05-18_pv_road_railway_brandenburg_kreise_epsg32633.csv",
         wind="raw/area_potential/2021-05-18_wind_brandenburg_kreise_epsg32633.csv",
         kreise="raw/lookup_table_brandenburg_kreise.csv",
-        assumptions="raw/scalars.csv",
+        assumptions="raw/scalars_base.csv",
         script="scripts/prepare_re_potential.py"
     output:
         directory("results/_resources/RE_potential/")
```
MaGering, line 149: I don't know if this is already planned / intended here, but this would ensure that the correct scalars are selected via wildcards:

```suggestion
        assumptions="raw/scalars_{range}.csv",
        script="scripts/prepare_re_potential.py"
    output:
        directory("results/_resources/{range}/RE_potential/")
    wildcard_constraints:
        range=("base|high|low")
```

 However, if we were to build on our current scalars structure, then the wildcards would have to be extended by the specification of a year and hence adjusted to:

```suggestion
        assumptions="raw/scalars_{range}_{year}.csv",
        script="scripts/prepare_re_potential.py"
    output:
        directory("results/_resources/{range}_{year}/RE_potential/")
    wildcard_constraints:
        range=("base|high|low"),
        year=("2040|2050")        
```

Just a thought without any particular change request.

SabineHaas: Thank you @MaGering! 
The thing is, that using the different csvs does not have any effect on the input data to the simulations, as the renewable potential is not copied to `_resources/scal_base.csv`. 
I would add the potential automatically in `prepare_scalars.py` using `results/_resources/scal_power_potential_wind_pv.csv` as input. Do you agree @jnnr?

SabineHaas: @MaGering I will include that, giving you credit, in #176 after the release. thanks again!

jnnr (approved):

PR #177, opened 2022-03-14
Feature/improve installation docs

### LINKED ISSUE #162: virtualenv instead of conda env not documented
You can actually use virtualenv and install via poetry install, but this is not documented.

MaGering:
Resolves #162 

With this PR

1. the advise of the possibility of using other virtual environments such as `virtualenv`,
2. the revision of the description for the installation of the docs and
3. the fix of wrong indentation in the docstrings of `prepare_re_feedin.py`

are included in the docs.

### DIFF
--- a/docs/docs_requirements.txt
+++ b/docs/docs_requirements.txt
@@ -1,3 +1,4 @@
+# Needed for installation of docs
 sphinx
 sphinx-rtd-theme>=0.5.1
 sphinxcontrib_bibtex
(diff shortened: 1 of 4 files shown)


henhuy (commented):
Added a minor fix for poetry (is installed from top level folder, instead of docs folder).
Also I have to say that I get a very long list of warnings when running `make html`.
Maybe you don't get them, as running command twice, warnings are not shown again.
You can deactivate caching via ` make html SPHINXOPTS=-E` - then you get all warnings again. Sorry!

docs/docs_requirements.txt
```diff
@@ -1,3 +1,4 @@
+# Needed for installation of docs
```
jnnr, line 1:
```suggestion
# Duplicates the docs requirements in pyproject.toml, but is necessary for readthedocs.
```

PR #181, opened 2022-03-28
Scenario handling for time series

jnnr:
This PR clarifies the `scenario_key` for timeseries. 
* With this PR, it assumes similar values as e.g.'Base 2050' (necessary for heat demands). If a timeseries is not varied in different scenarios, `scenario_key` is 'ALL'. 
* The information about the underlying year does not have to be represented in `scenario_key`, as it is already in `timeindex_start`. The filters in the scenario specifications are adapted accordingly.
* A bug in the selection of the regions. prepare_heat_loads.py has been fixed.

### DIFF
--- a/scenarios/base-2050-high_capacity_cost.yml
+++ b/scenarios/base-2050-high_capacity_cost.yml
@@ -105,4 +105,7 @@ paths_timeseries:
 
 filter_timeseries:
   scenario_key:
-    ts_2017
+    - "Base 2050"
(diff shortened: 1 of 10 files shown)


SabineHaas (approved):
Looks good to me! 
Only one small thing: the ROR time series need update, as well (they still have keys like ts_2017).

I did not fully test the PR, only let it run until `_resources` were build.

PR #187, opened 2022-04-01
Rebuild scenarios

jnnr:
Add some more scenarios using the centrally defined model structure #185 and the split data structure #186.


SabineHaas (commented):
That's a nice surprise for me, thank you @jnnr :) 

- We need use biomass
- I think we need a scenario_key for the emissions 80 95 100. Now, I don't see where the information comes from. I suggest `reduction_95`
- is there a difference between ` timeindex_start` and `datetimeindex: start:` ?

PR #189, opened 2022-04-04
Update oemoflex

jnnr:
Update oemoflex 
which requires updated oemof.tabular
which requires oemof.solph 0.4.

To test, run `snakemake -j1 results/example_base/optimized`


jnnr:
Got an error when constructing the links.

```
ERROR: Rule failed when generating expression for constraint LinkBlock.flow1
    with index ("<oemof.solph.network.bus.Bus: 'BE-electricity'>",
    "<oemof.tabular.facades.Link: 'BE-BB-electricity-transmission'>", 0):
    TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
ERROR: Constructing component 'LinkBlock.flow1' from data=None failed:
    TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

henhuy:
But this is related to your link fix and should be fixed...
Did you run `poetry install` to install correct oemof.solph version?

jnnr:
Thanks for updating this, @henhuy! I tested it with the newest data and a fresh environment and got the same results as with the dev branch (up to some differences due to the optimization being underdetermined).

As we decided to switch off the constraints for unidirectional flow in oemof.tabular already, this had to be reverted: [669a1a7](https://github.com/rl-institut/oemof-B3/pull/189/commits/669a1a70f29bf98e40116edbbca549699af02056).

PR #204, opened 2022-05-02
Implement new set of colors

monika-o:
On this branch, the colors of the plots have been revised to enhance visibility and intuitiveness.

### DIFF
--- a/oemof_b3/config/colors.csv
+++ b/oemof_b3/config/colors.csv
@@ -1,67 +1,67 @@
 ,Color
-PV,#bcbd22
+PV,#feff01
 Wind on,#1f77b4
-Wind off,#17becf
+Wind off,#73baec
 Biomass ST,#2ca02c
 Biomass GT,#007000
-Hydro reservoir,lightseagreen
-Hydro ROR,turquoise
+Hydro reservoir,#3c52f7
+Hydro ROR,#2837a9
 
-H2 GT,magenta
-CH4 GT,#d62728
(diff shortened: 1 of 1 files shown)


scripts/plot_scalar_results.py
```diff
@@ -487,6 +506,9 @@ def plot_demands(carriers):
     plot_flow_out_multi_carrier(CARRIERS)
     plot_demands(CARRIERS)
 
+    fig = draw_standalone_legend(colors_odict)
+    plt.savefig(os.path.join(target, "legend.png"))
```
jnnr, line 510:
```suggestion
    standalone_legend = False
    if standalone_legend:
        fig = draw_standalone_legend(colors_odict)
        plt.savefig(os.path.join(target, "legend.png"))
```

scripts/plot_dispatch.py
```diff
@@ -141,8 +145,40 @@
...
+            for key, value in simple_labels_dict.items():
+                if value[0] in labels and value[1] in labels:
+                    labels = [
+                        key
+                        if item == value[0]
+                        else "_Hidden"
+                        if item == value[1]
+                        else item
+                        for item in labels
+                    ]
```
jnnr, line 175: Could you make this a function or write an inline comment to make it easier to understand?

jnnr (commented):
Looks good! If you could make the relabeling step a bit cleaner, this would help. After that, you can merge.

PR #208, opened 2022-05-03
Removed wildcard from logfile in prepare_heat_demand rule

### LINKED ISSUE #206: WildcardError in prepare_heat_demand rule
When running `snakemake -j1 results/2050-el_eff/optimized` following error appears:
```WildcardError in line 121 of /home/local/RL-INSTITUT/hendrik.huyskens/Dokumente/RLI/oemof-B3/Snakefile:
Wildcards in params cannot be determined from output files. Note that you have to use a function to deactivate automatic wildcard expansion in params strings, e.g., `lambda wildcards: '{test}'`. Also see https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#non-file-parameters-for-rules:
'scenario'
```
Rewriting logfile parameter without wildcard `scenario` solves the issue - but I'm unsure if this is okay for you?
```
params:
        logfile="logs/heat_demand.log"
```

henhuy:
Fixes #206

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -129,7 +129,7 @@ rule prepare_heat_demand:
         scalars="results/_resources/scal_load_heat.csv",
         timeseries="results/_resources/ts_load_heat.csv",
     params:
-        logfile="logs/{scenario}.log"
+        logfile="logs/heat_demand.log"
     shell:
         "python scripts/prepare_heat_demand.py {input.weather} {input.distribution_hh} {input.holidays} {input.building_class} {input.scalars} {output.scalars} {output.timeseries} {params.logfile}"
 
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -129,7 +129,7 @@ rule prepare_heat_demand:
         scalars="results/_resources/scal_load_heat.csv",
         timeseries="results/_resources/ts_load_heat.csv",
     params:
-        logfile="logs/{scenario}.log"
+        logfile="logs/heat_demand.log"
```
jnnr, line 132:
```suggestion
        logfile="logs/prepare_heat_demand.log"
```

jnnr (approved):

PR #235, opened 2022-06-10
Refactored labels and colors

henhuy:
This PR introduces support for multiple label files.
Label files are now stored in folder _config/labels_ and can be selected via key `labels: de` in _settings.yml_.
Keys of label files now always refer to raw carriers (not translated carriers)!

As result of refactoring labels, I refactored colors as well, as they referred to translated labels.
For consistency, I also refactored colors in to yaml file.

Please double check if your scenarios are working, as changing labels and colors was a huge manual task (typing/spelling errors are most likely possible!)


jnnr (commented):
Thanks for refactoring, @henhuy! Looks good and more clear. Will test it soon. But first, one comment: Colors where defined as csv and loaded as an OrderedDict for a reason: The order of the colors is used by the dispatch plots to define the stacking order. IIRC yaml files do not support a defined order. Can you check that please?

jnnr:
> Starting from python >=3.6 all dicts are ordered by default. Thus, pyyaml should preserve order (see also https://stackoverflow.com/a/21912744/5804947). Does it not?

Could be. As I remember, there was no guarantee for python dicts to be ordered, even if they happen to be ordered.

PR #236, opened 2022-06-10
Scenario labels can be set for joined plots

henhuy:
`plot_scalar_results` looks for key `label` in related scenario yaml and replaces plot label, if it can find one.
If it fails to find a label, original scenario name is taken. 
Added label to scenario _2050-el_eff_ for testing purposes.
Tested with `snakemake -j1 results/joined_scenarios/all-optimized/joined_plotted`

### DIFF
--- a/scenarios/2050-el_eff.yml
+++ b/scenarios/2050-el_eff.yml
@@ -4,6 +4,7 @@
 # build_datapackage
 
 name: 2050-el_eff
+label: Efficiency-Scenario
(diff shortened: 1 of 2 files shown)


SabineHaas (approved):
Already tested it - works fine! 

Thanks a lot @henhuy  :grinning:

jnnr (approved):
Thanks, @henhuy! Looks good.

PR #246, opened 2022-08-09
Replace df.append with pandas.concat

### LINKED ISSUE #180: FutureWarning in data_processing
```
FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
df_stacked = df_stacked.append(df_stacked_column, ignore_index=True)
```

jnnr:
Fixes #180.

### DIFF
--- a/oemof_b3/tools/data_processing.py
+++ b/oemof_b3/tools/data_processing.py
@@ -606,7 +606,7 @@ def expand_regions(scalars, regions, where="ALL"):
 
         regionalized["region"] = region
 
-        sc_with_region = sc_with_region.append(regionalized)
+        sc_with_region = pd.concat([sc_with_region, regionalized])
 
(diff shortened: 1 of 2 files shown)


SabineHaas (commented):
The concatenation looks fine to me! I didn't test the code though; shall I do that? 
Have you checked the result of the concatenation, e.g. axis parameter is correct?

jnnr:
> The concatenation looks fine to me! I didn't test the code though; shall I do that? Have you checked the result of the concatenation, e.g. axis parameter is correct?

Have checked it with a small example, but not the whole code. If you want, give it a try. Otherwise, I can do it later, too.

SabineHaas (approved):
The concatenated data frames look good. 
I have checked all apart from `_df_wo_cc` in `create_empty_scalars.py` as this does not have a snakemake rule.

PR #247, opened 2022-08-10
Save logs next to results

jnnr:
To document results, logs telling about how the result has been created should be found in the same directory as the result.


jnnr:
@SabineHaas: I will test this before merging.

SabineHaas (approved):

PR #270, opened 2022-10-11
Include description of results in description of  model pipeline

jnnr:
Model pipeline (explaining the rules) and model structure follow the same logic. This is a proposal to integrate them. The different sections structure the snakemake rules by different categories (prepare_resources, build_datapackage etc.). Each of the categories have similar output (resources format, preprocessed datapackages, plots). The general structure of that output is explained in the text copied from model_structure. More detailled info on the rules can be obtained by clicking on a specific rule.

[readthedocs-build](https://oemof-b3.readthedocs.io/en/feature-restructure-docs-model-structure/)

Please let me know what you think!

TODOs:

* [x] Move all important text parts to model pipeline
   * [x] raw data
   * [x] ..
* [x] Delete model structure


MaGering:
Thank you @jnnr for the restructuring. I think your changes simplify the orientation in the docs.

This means that "Model pipeline" and "Model structure" will be merged into one chapter which is not yet implemented in this draft, right?

jnnr:
> Thank you @jnnr for the restructuring. I think your changes simplify the orientation in the docs.
> This means that "Model pipeline" and "Model structure" will be merged into one chapter which is not yet implemented in this draft, right?

Thanks for your feedback! Yes, they will be merged. I will delete "Model structure"

jnnr:
Raw data does not fit the logic of the sections, which are ordered by rule type. Proposal: Change the sections like this
* Raw data
* Resources
  * rules `prepare_`
* Preprocessed Datapackages
   * rule `build_datapackage`
* Optimized Energysystems
   * rule `optimize`
* Postprocessed Datapackages
   * rule `postprocessed`
* Plots
   * ruled `plot_`

MaGering:
> Raw data does not fit the logic of the sections, which are ordered by rule type. Proposal: Change the sections like this
> 
>     * Raw data
> 
>     * Resources
>       
>       * rules `prepare_`
> 
>     * Preprocessed Datapackages
>       
>       * rule `build_datapackage`
> 
>     * Optimized Energysystems
>       
>       * rule `optimize`
> 
>     * Postprocessed Datapackages
>       
>       * rule `postprocessed`
> 
>     * Plots
>       
>       * ruled `plot_`

I think that's good. (With only one addition: Instead of "Plots" to write "Visualization", because there also the table scripts are included - this will come automatically with the merge of dev).

Issue #258, opened 2022-09-05
Data processing function use print, not logger


jnnr:
As the title says, many functions in tools.dataprocessing use print. Should be replaced with logger.info or logger.warning.

jnnr:
Could you take this one, @monika-o?

monika-o:
Yes!

jnnr:
Closed due to #273

PR #273, opened 2022-10-13
Logger in data_processing.py

monika-o:
This little pull request introduces logging to tools/data_processing.py (see #258).

### DIFF
--- a/scripts/build_datapackage.py
+++ b/scripts/build_datapackage.py
@@ -56,6 +56,7 @@ from oemof_b3.config import config
 
 logger = logging.getLogger()
 
+
 def update_with_checks(old, new):
     r"""
     Updates a Series or DataFrame with new data. Raises a warning if there is new data that is not
--- a/scripts/build_datapackage.py
+++ b/scripts/build_datapackage.py
@@ -26,7 +26,6 @@ Explanations about the structure of the preprocessed datapackage can be found in
 `docu <https://oemof-b3.readthedocs.io/en/latest/index.html>`_.
 
 """
-import logging
 import sys
 import os
 from collections import OrderedDict
@@ -50,11 +49,8 @@ from oemof_b3.tools.data_processing import (
     unstack_timeseries,
     expand_regions,
     save_df,
-    initLogger,
 )
 from oemof_b3.config import config
-
-logger = logging.getLogger()
 
 
 def update_with_checks(old, new):
@@ -248,9 +244,7 @@ if __name__ == "__main__":
 
     destination = sys.argv[2]
 
-    logfile = sys.argv[3]
-    logger = config.add_snake_logger(logfile, "build_datapackage")
-    initLogger(logger)
+    logger = config.add_snake_logger("build_datapackage")
 
     scenario_specs = load_yaml(scenario_specs)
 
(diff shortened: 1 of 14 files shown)


monika-o:
I've made the changes similar to the logger implementation in scripts/build_datapackage.py. I tried to test it by running the scenario 2050-100-el_eff up to the preprocessing step. This didn't produce an error, but I didn't see any of the statements in the logfile (2050-100-el_eff.log), so I am not sure whether the logging didn't work or whether it just hasn't been called. @jnnr Do you have an idea how to test it, or are the changes so small and clear that we can just merge?

jnnr:
Thank you! The location of the logfiles has changed with #247 to save the logfiles next to the results. I would check if the files logging the data processing steps can be found in results/_resources.

monika-o:
I think I have found the correct logfile, but I've seen now that there are some lines that I have overlooked. I will notify you when I am done.

monika-o:
Here is an update: Some of the logging statements are indeed called when testing it, and they are printed on the terminal. The problem is that they are not saved to the logfile (2050-100-el_eff.log, and also not to any other logfile). I am now trying to figure out the reason for this and would be glad about your ideas if you have some.

monika-o:
The reason is that the functions in data_processing.py use <RootLogger root (INFO)> as logger, whereas the script build_datapackage.py uses the local logger <Logger build_datapackage (INFO)> (and the same issue most probably occurs with optimize.py).

monika-o:
I see two ways to solve this:
1. Simply passing the logger as a parameter to the function (in a total of 6 functions in which we use logging) or
2. Creating a new logger for data processing. This logger, however, must know the location of the logfile (i.e. the scenario) and I didn't find out yet how to access that information.

jnnr:
Thanks for your approach. Looks good, but will break if the naming convention of naming the logger same as the script is not obeyed. 

I found this: https://stackoverflow.com/questions/26405841/python-logging-from-library-module

scripts/build_datapackage.py
```diff
@@ -243,6 +245,7 @@ def calculate_emission_limit(
 
     logfile = sys.argv[3]
     logger = config.add_snake_logger(logfile, "build_datapackage")
+    initLogger(logger)
```
jnnr, line 248: Could this function be called from within config.add_snake_logger? This would avoid a lot of code duplication in the scripts.

henhuy:
Wasn't as easy as I expected.
Now logger is added in data_preprocessing.py.
Name of logfile is read from input parameters and MUST end with ".log" to be detected.

One off-topic commit: Had to fix pre-commit tag (master/main not possible any more - must be tagged to specific version)

jnnr (approved):
Thanks for all your work for this, @monika-o and @henhuy!

I did not expect it to be a problem that uncommon to lead to so many different workarounds. 

To put an end to this, I would like to accept your solution, @henhuy. The only difference to what I originally had in mind is that the logger in dataprocessing has the category "data_processing" and does not state the script from which it is called. But I am ok with it.

henhuy:
> Thanks for all your work for this, @monika-o and @henhuy!
> 
> I did not expect it to be a problem that uncommon to lead to so many different workarounds.
> 
> To put an end to this, I would like to accept your solution, @henhuy. The only difference to what I originally had in mind is that the logger in dataprocessing has the category "data_processing" and does not state the script from which it is called. But I am ok with it.

Oh no. Had this implemented at one point, but thought it would be better to see module name... :(
But as this issue already took so long, I also would close it here...

monika-o:
I would also be fine with it if you close this pull request now.

PR #275, opened 2022-10-24
Feature/improve docstrings

MaGering:
In this PR the docstrings of the python files within `oemof-B3` `scripts` directory are revised.


MaGering:
Is the following still true?

[Please note that this script is not integrated into the pipeline, yet.](https://github.com/rl-institut/oemof-B3/blob/2f4e8183ac39c730a71f4c9ffff4a1b6d03d78ef/scripts/prepare_re_potential.py#L32)

scripts/plot_storage_levels.py
```diff
@@ -19,8 +19,8 @@
 -------------
 The script creates storage level plots.
 
-The static plots are saved as files (png, pdf, etc.) and the interactive plotly plots as html-files
-in a new directory called plotted.
+The static plots are saved as png-files and the interactive plotly plots as html-files
```
jnnr, line 22: Same as before. The file extension is defined in config.settings.general.plot_filetype

jnnr (approved):
Thanks for the corrections. I found two more. With them fixed, you can merge.

scripts/plot_dispatch.py
```diff
@@ -20,8 +20,8 @@
 -------------
 The script creates dispatch plots based on plot_dispatch and plot_dispatch_plotly
 functions in oemoflex.
-The static plots are saved as pdf-files and the interactive plotly plots as html-files
-in a new directory called plotted.
+The static plots are saved as png-files and the interactive plotly plots as html-files
```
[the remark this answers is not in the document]
MaGering: I changed that with commit https://github.com/rl-institut/oemof-B3/pull/275/commits/f11f70de0f1b1565352e3c0863dccee1ff44dcb6. Is that what you had in mind?

scripts/plot_storage_levels.py
MaGering: See above. Also done in other docstrings of scripts importing `plot_filetype`.

jnnr (approved):

PR #276, opened 2022-10-24
Keep multiindex level var_name

### LINKED ISSUE #262: plot_all_examples not working
**Updated error report:**

Plotting with `plot_scalar_results.py` does not work for examples and single scenarios. There are the following errors: 

1. capacities have been removed from results. This leads to an error running
```
    plot_capacities()
```
We will drop this function as the capacities are input parameters, not results.

2. These plotting functions do not work with single scenarios, we have to find out why:
```
    subplot_invest_out_multi_carrier(CARRIERS_WO_CH4)
    subplot_storage_invest_multi_carrier(CARRIERS_WO_CH4)
    subplot_flow_out_multi_carrier(CARRIERS_WO_CH4)
    subplot_demands(CARRIERS)
    subplot_energy_usage_multi_carrier(CARRIERS)
```

------------------------------------
**Original error report:**

I can't run

```
snakemake -j1 plot_all_examples
```
due to `FileNotFoundError`:
```
[Errno 2] No such file or directory: 'scenarios/example_base.yml'
```
Is there something one has to do in order to get the examples running (copying a file from A to B)? If yes, it is not mentioned in the [docs](https://oemof-b3.readthedocs.io/en/latest/examples.html) so far and the label needs to change from `bug` to `documentation`.

### LINKED ISSUE #271: Plot scalars stopped working for single scenarios
plot_scalars was adapted to work with pd.MultiIndex in case of several scenarios. With only one scenario, it fails. Leads to #262

MaGering:
Resolves #271 

Plotting removes all multiindex levels that have only 1 value. This PR allows to define levels that should be kept and uses this to keep "var_name".

Note that issue https://github.com/rl-institut/oemof-B3/issues/278 persists at the time this PR was created. In order to test this PR the following functions need to be commented out:
```
    plot_capacities()
    subplot_invest_out_multi_carrier(CARRIERS_WO_CH4)
    subplot_storage_invest_multi_carrier(CARRIERS_WO_CH4)
    subplot_flow_out_multi_carrier(CARRIERS_WO_CH4)
    subplot_demands(CARRIERS)
    subplot_energy_usage_multi_carrier(CARRIERS)
```


MaGering:
@jnnr, the fix does not work yet. So this PR is not ready for a review yet.

MaGering:
> @jnnr, the fix does not work yet. So this PR is not ready for a review yet.

Okay I could locate the bug. It was due to empty data in `plot_capacities` function. In PR https://github.com/rl-institut/oemoflex/pull/74 we removed capacities from the results since it is an input parameter, not a result.

We will drop `plot_capacities` with the changes that resolve issue https://github.com/rl-institut/oemof-B3/issues/278.

This PR is ready for review @jnnr. To do this you'll need to comment `plot_capacities` out together with the subplot functions (see my comment at the top).

jnnr (approved):

PR #281, opened 2022-11-08
Features/set name attr

### LINKED ISSUE #137: Scalars: name stays empty if region is not 'ALL'
If `region` is not `ALL` but a region, `name` is not set automatically. In #123 I just added the `name` manually in the scalars. 

We should e.g.
- set the `name` also for parameters which have a `region` different from `ALL`
- require that `name` is filled in for all parameters in the scalars

Maybe there's also another solution.

MaGering:
Resolves #137

With this PR the attribute `name` is checked within the script `data_processing.py`.

With the implemented function it is ensured that the name is

1. empty if region is 'ALL'
2. set (according to convention) where name is empty and region is fixed
3. checked for all values that are not None and where region is fixed


jnnr (commented):
Thanks for your work! Well written and well-documented - looks really good.

One complaint though: The function does many things at a time and is therefore hard to understand. I have some ideas to make it simpler:

* Do the checking and setting after the regionalization, so your function does not have to distingiush betweenwith_region and wo_region.
* Define a function get_name(region, carrier, tech) once and use it in line 719 and in your check_name/set_name function
* Could use https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_series_equal.html to compare two pd.series for "name".

oemof_b3/tools/data_processing.py
```diff
@@ -592,23 +709,28 @@ def expand_regions(scalars, regions, where="ALL"):
...
-        return sc_with_region
-
-    for region in regions:
-        regionalized = sc_wo_region.copy()
-
-        regionalized["name"] = regionalized.apply(
-            lambda x: "-".join([region, x["carrier"], x["tech"]]), 1
-        )
+    # REGIONALIZATION
+    if not sc_wo_region.empty:
```
jnnr, line 713: Why did you change the if statement? Before, both cases where covered. Now, if sc_wo_region is empty, you will run into an error when trying to return sc_with_region.

oemof_b3/tools/data_processing.py
```diff
@@ -565,6 +567,121 @@ def aggregate_timeseries(df, columns_to_aggregate, agg_method=None):
     return df_aggregated
 
 
+def prepare_attr_name(sc, regions, overwrite):
```
jnnr, line 570: This function should only set the name according to region, carrier, tech provided in the scalars. No need to pass regions.

jnnr (commented):
Thanks for your work! 
I have to comments. Hope they help to make it even simpler!

oemof_b3/tools/data_processing.py
MaGering: No, I don't because the functionality remains same returning `sc_with_region` in line 737 if no regionalization happens (if statement line 715 - 735). I've tested it with data which has region specified. 

While I was doing that, I caught another error, which results from an empty DataFrame in case of no data with unset name. I made the corresponding changes in commit https://github.com/rl-institut/oemof-B3/pull/281/commits/84b67813e00e2326be37e65c4a98937aede07b8d.

jnnr: Ah, you are right, sorry. I missed line 737.

oemof_b3/tools/data_processing.py
MaGering: You mean `set_name` function, don't you? Unfortunately, I don't know how to implement this so that it is simplified and not more complicated. Could you maybe write a suggestion?

MaGering:
> Hi @MaGering, I had something like this in mind (can be adapted, just a sketch and did not test it):
> 
> ```
> def get_name(region, carrier, tech):
>     r"""
>     This is oemof-b3's-naming convention.    
>     """
>     return f"{region}-{carrier}-{tech}
> 
> def get_name_for_df(df):
>     # TODO: could assert that df has region carrier tech as columns.
>     return df.apply(lambda x: get_name(x["region"], x["carrier"], x["tech"]), 1)
> 
> def check_name(df):
>     name_generated = get_name_for_df(df)
>     name_as_given = df["name"]
>     
>     diff = compare_scalar_data(name_given, name_generated)
> 
>     if diff:
>         raise Warning(f"There is a diff {diff}")
> 
> 
> def set_name(df):
>     try:
>         check_name_df(df)
>     except Warning as w:
>         raise Warning("Name will be adapted"}   
>  
>     name_generated = get_name_for_df(df)
>     df.name = name_generated
> 
>     return df
>     
> ```

Thanks for these suggestions. They were really helpful! Done with commit https://github.com/rl-institut/oemof-B3/pull/281/commits/595be9571ebc96b80f14a392cb66e298a7cfa033.

MaGering:
Ready for your review @jnnr!

jnnr (commented):
Tested and got 
* SettingWithCopyWarning
* A difference in additional scalars. The new functions introduced the name `TOTAL-emission-constraint` in three rows.

oemof_b3/tools/data_processing.py
```diff
@@ -599,6 +599,172 @@ def aggregate_timeseries(df, columns_to_aggregate, agg_method=None):
...
+            "from the convention (<region>-<carrier>-<tech>). \n"
+            "We expected but could not find the following name(s): "
+            f"{expected_names}."
+        )
+        if overwrite:
+            logger.warning(
+                "The names will be overwritten with names following the convention"
+            )
+
+    def set_name(df):
```
jnnr, line 705: Better pass overwrite explicitly

oemof_b3/tools/data_processing.py
```diff
@@ -599,6 +599,172 @@ def aggregate_timeseries(df, columns_to_aggregate, agg_method=None):
...
+        Returns
+        -------
+        df : pd.DataFrame
+            DataFrame in oemof-B3-resources format and formatted name.
+
+        """
+        if not df["name"].isnull().values.all():
+            check_name(df)
+
+        if overwrite or df["name"].isnull().values.all():
```
jnnr, line 726:
```suggestion
        all_empty = df["name"].isnull().values.all()
        if not all_empty:
            check_name(df)

        elif all_empty or overwrite:
```

oemof_b3/tools/data_processing.py
MaGering: Done with commit https://github.com/rl-institut/oemof-B3/pull/281/commits/48e7ab85c40bc993f72d4fea45c8fb4ade5522e6.

MaGering:
> Tested and got
> 
>     * SettingWithCopyWarning

Commit https://github.com/rl-institut/oemof-B3/pull/281/commits/46a2be03594bd93480ac41e83ee99abba7894670 did not solve the problem as suggested in the console. Exactly the same warning appeared in the `pandas` datapackage instead of  in the `data_processing.py` script.

I kept commit https://github.com/rl-institut/oemof-B3/pull/281/commits/46a2be03594bd93480ac41e83ee99abba7894670 anyway, as it is preferable to the previous approach of overwriting.

I could finally solve the problem using a copy of the Dataframe `df`. With the copy I am able to overwrite the values without the warning. Afterwards I overwrite the original `df` with the copy (also without a warning). See commit https://github.com/rl-institut/oemof-B3/pull/281/commits/8d083d78c702c59c4db1b6ae129577c6e6b5daf7. I tested this and compared the outcome with meld.

jnnr:
> Tested and got
>     * A difference in additional scalars. The new functions introduced the name `TOTAL-emission-constraint` in three rows.

We found out that this has no influence on the results as the name is not processed for the emission constraints. I opened an issue that reminds us of that.

jnnr (approved):
Tested it by building the datapackage and comparing to previous results.

PR #285, opened 2022-11-24
Make stack unstack of scalars reversible

jnnr:


tests/test_data_processing.py
```diff
@@ -492,6 +494,27 @@ def test_stack_unstack_on_example_data():
     assert pd.testing.assert_frame_equal(df, df_unstacked) is None
 
 
+def test_stack_unstack_scalars_on_example_data():
```
jnnr, line 497:
```suggestion
def test_unstack_stack_scalars_on_example_data():
```

jnnr:
Hi @MaGering, can you test if these changes change anything in the results when running the snakemake pipeline?

MaGering:
> Hi @MaGering, can you test if these changes change anything in the results when running the snakemake pipeline?

I compared the results of a scenario on dev with the results of the same scenario on this branch and they look the same to me.

oemof_b3/tools/data_processing.py
```diff
@@ -142,6 +142,26 @@ def load_b3_scalars(path, sep=";"):
...
+        path of input file of csv format
+    sep : str
+        column separator
+
+    Returns
+    -------
+    df : pd.DataFrame
+        DataFrame with unstacked scalars
+    """
+    df = pd.read_csv(path, sep=sep, index_col=[0, 1, 2, 3, 4, 5], header=[0, 1])
```
MaGering, line 161: Where does this format come from?

oemof_b3/tools/data_processing.py
```diff
@@ -142,6 +142,26 @@ def load_b3_scalars(path, sep=";"):
     return df
 
 
+def load_unstacked_b3_scalars(path, sep=";"):
```
MaGering, line 145: Have you tested this function? So far it is not used anywhere right?

jnnr:
> > Hi @MaGering, can you test if these changes change anything in the results when running the snakemake pipeline?
> 
> I compared the results of a scenario on dev with the results of the same scenario on this branch and they look the same to me.

Thank you. In this case I will merge this.

MaGering (commented):
Thanks for your changes! I was wondering where they are used and have two more specific questions about them.

oemof_b3/tools/data_processing.py
jnnr: You are right. It is not used and not tested. I guess I needed it when playing around with the functions, and it may be used sometimes. But currently not. Better delete it?

oemof_b3/tools/data_processing.py
jnnr: This is how the unstacked scalars look like when taking B3-scalars and apply unstack_var_name.

PR #287, opened 2022-11-28
Feature/dataprocessing classes

jnnr:
Replaces the ScalarProcessor and simplifies working with oemof-B3 data.


jnnr:
Could reproduce the previous data when running `snakemake -j1 results/_resources/scal_cost_efficiencies.csv`

PR #288, opened 2022-12-01
Added script to download resources

henhuy:
Added script `download_resources` which downloads resources from OEP and saves them in folder `results/_resources`.
TODOS:
- [x] Add all resources to list of resources in `download_resources.py`
- [x] Add rule to snakemake

@jnnr could you complete TODOS above, as you might know better all resources and related tables on OEP. and you are more familiar with snakemake...?

- [x] Downloaded resources are separated by `,`, but are expected `;`
- [x] String entries are enclosed with `""`
- [x] A new column "id" occurs


henhuy:
Download of `raw_data` could be implemented easily by using function `download_table_from_OEP` from `oemof_b3.tools.oep`

jnnr:
Downloading raw data is much easier to implement into the pipeline than resources. This is because there is no rule for raw data yet, but there are many rules for creating resources.

jnnr:
Found a solution for the above problem: Moving all rules that prepare resources in a submodule that can be included or not depending on a value in the settings. #290

jnnr:
Almost there. Still missing:
* [  ] A list of all resources. Maybe read from the snakemake rules outputs?
* [ ] Pass the target directory to script from snakemake

jnnr:
The resources can be downloaded, but there are differences to the ones created locally. First, the delimiter is different. Second, entries are in "".

jnnr (commented):
Thanks for your help, @henhuy. I finished the open todos and we are almost ready. But the downloaded csv files have some formatting difference, see the todos above. Can you fix them?

PR #290, opened 2022-12-05
Modularize snakemake rules

jnnr:
Modularizes snakemake rules to make it easier switching off the whole prepare-resources pipeline in case one wants to download resources from oep.


jnnr:
Hi @MaGering, I wanted to give you the chance to have a quick look here so you can follow what is going on. If you have feedback, it is welcome, but you do not have to check every line of code. I just copied the rules to submodules.

PR #307, opened 2023-01-09
Features/update create empty scalars

MaGering:
With this PR a bug in create_empty_scalars.py is fixed.
Besides these changes the script is cleaned up.

The saving function is improved to allow a potential future loop over scenarios. If there is already an empty scalars file in the destination path missing lines will be updated.

### DIFF
--- a/scripts/create_empty_scalars.py
+++ b/scripts/create_empty_scalars.py
@@ -17,15 +17,16 @@ Description
 The script creates an empty DataFrame for scalar data that serve as a template for input data.
 """
 import sys
-
+import os
 import pandas as pd
+
 from oemoflex.model.datapackage import EnergyDataPackage
 
 from oemof_b3.config.config import load_yaml, settings
 from oemof_b3.model import bus_attrs_update, component_attrs_update, model_structures
-
 from oemof_b3.tools.data_processing import (
     HEADER_B3_SCAL,
+    load_b3_scalars,
     format_header,
     sort_values,
     save_df,
@@ -117,6 +118,25 @@ def add_wacc(sc):
     return sc
 
 
+def save_empty_scalars(sc, path):
+
+    if os.path.exists(path):
+        all_sc = load_b3_scalars(path)
+
+        if all_sc.empty:
+            all_sc = sc
+        else:
+            all_sc = all_sc.append(sc, ignore_index=True)
+            all_sc.index.name = sc.index.name
+
+    else:
+        all_sc = sc
+
+    all_sc.drop_duplicates(inplace=True)
+
+    save_df(all_sc, path)
+
+
 if __name__ == "__main__":
     scenario_specs = sys.argv[1]
 
@@ -155,4 +175,4 @@ if __name__ == "__main__":
 
     empty_scalars = sort_values(empty_scalars)
 
-    save_df(empty_scalars, destination)
+    save_empty_scalars(empty_scalars, destination)
--- a/scripts/create_empty_scalars.py
+++ b/scripts/create_empty_scalars.py
@@ -34,6 +34,34 @@ from oemof_b3.tools.data_processing import (
 NON_REGIONAL = settings.create_empty_scalars.non_regional
 
 
+def get_edp_from_scenario(scenario_specs):
+    model_structure = model_structures[scenario_specs["model_structure"]]
+
+    # setup empty EnergyDataPackage
+    datetimeindex = pd.date_range(
+        start=scenario_specs["datetimeindex"]["start"],
+        freq=scenario_specs["datetimeindex"]["freq"],
+        periods=scenario_specs["datetimeindex"]["periods"],
+    )
+
+    # setup default structure
+    edp = EnergyDataPackage.setup_default(
+        basepath=destination,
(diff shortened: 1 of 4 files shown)


MaGering:
Open question: 

I would drop this line because attributes disappear from scalars:
https://github.com/rl-institut/oemof-B3/blob/14656b04da0b4a85df8ece5bfde22986a34eb50e/scripts/create_empty_scalars.py#L53

scripts/create_empty_scalars.py
```diff
@@ -82,41 +101,57 @@ def expand_scalars(df, column, where, expand):
...
+            "type": "ALL",
+            "var_unit": "None",
+        },
+        ignore_index=True,
+    )
+
+    return sc
+
+
+def save_empty_scalars(sc, path):
```
jnnr, line 121: This function appends to an existing file. That can make it difficult to keep track of things. Do we really want to do this? If yes, it needs to be communicated more prominently (name of the function, docstrings). But first, think if we really need this!

scripts/create_empty_scalars.py
```diff
@@ -82,41 +101,57 @@ def expand_scalars(df, column, where, expand):
     return _df_wo_cc
 
 
+def add_wacc(sc):
```
jnnr, line 104: This function is very specific. Why not instead write a function that adds a line to the scalars and then pass the specific data to it?

jnnr (commented):
Thanks for cleaning up! Could not find the problem with the loop. Maybe you can post the error message or more info here?

jnnr:
> Open question:
> 
> I would drop this line because attributes disappear from scalars:
> 
> https://github.com/rl-institut/oemof-B3/blob/14656b04da0b4a85df8ece5bfde22986a34eb50e/scripts/create_empty_scalars.py#L53

This is dropped because these parameters are set by default (many of them are bus, from_bus, to_bus, carrier, tech e.g. parameters you do not want to specify again because they are given by the model structure). But you can change that or make that optional, of course.

scripts/create_empty_scalars.py
MaGering: Good idea! Thanks. Done something like this with commit https://github.com/rl-institut/oemof-B3/pull/307/commits/00606929469bd38bd30b0aa9fd660bf3e7085055.

MaGering:
> Could not find the problem with the loop. Maybe you can post the error message or more info here?

Did you implement the loop and run it?

You can find my changes I've made attached in `KeyError_loop_scenarios.zip` (`create_empty_scalars.py` and `create_empty_scalars.smk`).
The output of the console including the `KeyError` can be found in file `KeyError_loop_over_scenarios_component_attrs.txt`.
In `component_attrs_content_first_iteration.txt` you find the content of `component_attrs_update.yml` at the first loop step and in `component_attrs_content_first_iteration.txt` at the second loop step.

It is also interesting that the error persists even if I insert a copy of `component_attrs_update.yml` before the loop (`attr_of_comp = component_attrs_update.copy()`) and overwrite `component_attrs_update` in the loop with this copy. The copy then also contains the wrong data. What am I missing?

[KeyError_loop_scenarios.zip](https://github.com/rl-institut/oemof-B3/files/10385120/KeyError_loop_scenarios.zip)

scripts/create_empty_scalars.py
MaGering: Yes, this only makes sense if I get the loop implemented over the scenarios. Then there would be a single `empty_scalars.csv` which is updated with every iteration over the scenario. I have therefore left it in.

MaGering:
@jnnr: Should we consider this within milestone v0.0.2 or within v0.0.3?

MaGering:
> > Could not find the problem with the loop. Maybe you can post the error message or more info here?
> 
> Did you implement the loop and run it?
> 
> You can find my changes I've made attached in `KeyError_loop_scenarios.zip` (`create_empty_scalars.py` and `create_empty_scalars.smk`). The output of the console including the `KeyError` can be found in file `KeyError_loop_over_scenarios_component_attrs.txt`. In `component_attrs_content_first_iteration.txt` you find the content of `component_attrs_update.yml` at the first loop step and in `component_attrs_content_first_iteration.txt` at the second loop step.
> 
> It is also interesting that the error persists even if I insert a copy of `component_attrs_update.yml` before the loop (`attr_of_comp = component_attrs_update.copy()`) and overwrite `component_attrs_update` in the loop with this copy. The copy then also contains the wrong data. What am I missing?
> 
> [KeyError_loop_scenarios.zip](https://github.com/rl-institut/oemof-B3/files/10385120/KeyError_loop_scenarios.zip)

I fixed this with commit https://github.com/rl-institut/oemof-B3/pull/307/commits/997b5742b751c062c0b183add3d8a158dd385522. I'm loading `component_attrs_update` now from the [original yaml file](https://github.com/rl-institut/oemof-B3/blob/198a877adc1f0e98263e635791cd07904366356b/scripts/create_empty_scalars.py#L139-L141) instead via the model module. Via the model module it always was overwritten with the input from oemoflex no matter what I tried (cf. comment above).

PR #311, opened 2023-02-09
Write lp file to optimized dir if lp writing is activated

MaGering:
With this PR it is possible to activate the output of a lp-file.

### DIFF
--- a/oemof_b3/config/settings.yaml
+++ b/oemof_b3/config/settings.yaml
@@ -19,7 +19,7 @@ optimize:
   solve_kwargs:
     tee: True
     keepfiles: True
-  lp_writing: False
+  write_lp_file: False
   cmdline_options:
     AllowableGap: 0.01
   debug: true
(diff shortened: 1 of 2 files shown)


MaGering:
@jnnr: We do not have something like this already, do we?
I think it is handy. What do you think?

oemof_b3/config/settings.yaml
```diff
@@ -19,6 +19,7 @@ optimize:
   solve_kwargs:
     tee: True
     keepfiles: True
+  lp_writing: False
```
jnnr, line 22:
```suggestion
  write_lp_file: False
```

scripts/optimize.py
```diff
@@ -268,6 +268,12 @@ def get_additional_scalars():
         )
 
         with Timer(text="Solved the model.", logger=logger.info):
+            write_LP = config.settings.optimize.lp_writing
+            if write_LP:
+                m.write(
+                    os.path.join(optimized, "optimized.lp"),
+                    io_options={"symbolic_solver_labels": True},
+                )
```
jnnr, line 276:
```suggestion
            if config.settings.optimize.write_lp_file:
                m.write(
                    os.path.join(optimized, "optimized.lp"),
                    io_options={"symbolic_solver_labels": True},
                )
```

jnnr (approved):

PR #312, opened 2023-02-13
Add script and snakemake rule for raw download

jnnr:


MaGering (approved):
Thanks for these changes. Works perfectly. Just tested it. :)

Issue #318, opened 2023-02-21
Windows issue with snakemake is perhaps resolved


monika-o:
In the docs in the section "Snakemake on Windows" it says:

"When running snakemake with output files in subfolders on Windows with

snakemake -j<NUMBER_OF_CPU_CORES>

a MissingRuleException is raised. The process is unable to specify the output files in subfolders. This bug is an [open issue](https://github.com/snakemake/snakemake/issues/46) in snakemake."

This issue is closed now (since June 2021). It should be tested whether the problem has been resolved.

Stefanie08:
Comparing the run of the rule `snakemake -j<NUMBER_OF_CPU_CORES>` with the workaround of Issue https://github.com/rl-institut/oemof-B3/issues/26, with `snakemake snakemake -j<NUMBER_OF_CPU_CORES> --keep-target-files`, the same output is generated. Therefore, this issue can be closed.

PR #321, opened 2023-02-21
Remove Windows issue from docs

monika-o:
There has been an issue in Windows with snakemake (https://github.com/snakemake/snakemake/issues/46), but that issue is closed now. I have tested snakemake execution in Windows with several rules and it worked without error. Also pypsa-eur, which we cited as a source for the workaround, doesn't mention the issue anymore (https://pypsa-eur.readthedocs.io/en/latest/tutorial.html?highlight=windows#how-to-use-the-snakemake-rules). This PR therefore abolishes the explanation of the issue and the workaround in the docs. (corresponding issue: #318)


MaGering (approved):
Thanks for these changes @monika-o

PR #325, opened 2023-04-05
Features/add empty ts

MaGering:
With this PR it is possible to create empty time series in the directory `raw/time_series`.

Empty means in this case, that the time series contain either "0" or "nan" as values. This can be set via `create_empty_ts.ts_values` in `settings.yaml`.

The docs have been updated.

To test the new functionality run:
```
snakemake -j1 create_empty_ts
```
Before you need to [install oemof-B3 according to docs](https://oemof-b3.readthedocs.io/en/latest/getting_started.html) and download raw directory with:

```
snakemake -j1 download_raw_data
```

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -94,6 +94,7 @@ include: "snakemake_rules/postprocessing.smk"
 include: "snakemake_rules/visualization.smk"
 include: "snakemake_rules/oep_upload.smk"
 include: "snakemake_rules/create_empty_scalars.smk"
+include: "snakemake_rules/create_empty_ts.smk"
 
 # prepare settings locally or download it from OEP (not implemented yet)
 if config.settings.general.prepare_resources_locally:
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -143,13 +143,7 @@ Alternatively, to create just the output file or directory of one rule, run:
 Contributing to oemof-B3
 ========================
 
-Executing the rule
-
-::
-
-    snakemake -j1 create_empty_scalars
-
-will create empty scalars. The rule
+The rule
 
 ::
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -143,5 +143,21 @@ Alternatively, to create just the output file or directory of one rule, run:
 Contributing to oemof-B3
 ========================
 
+Executing the rule
+
+::
+
+    snakemake -j1 create_empty_scalars
+
+will create empty scalars. The rule
+
+::
+
+    snakemake -j1 create_empty_ts
+
+will create empty time series data.
+The empty scalars and time series data can be used to verify your energy system model in the preprocessing stage.
+
+
 You can write `issues <https://github.com/rl-institut/oemof-B3/issues>`_ to announce bugs or
 to propose enhancements.
--- a/docs/prepare_resources/create_empty_ts.rst
+++ b/docs/prepare_resources/create_empty_ts.rst
@@ -1,0 +1,7 @@
+.. _create_empty_ts_label:
+
+create_empty_ts
+===============
+
+.. automodule:: create_empty_ts
+   :members:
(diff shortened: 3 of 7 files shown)


chrwm (commented):
Hi @MaGering I had some problems getting oemof-b3 to run, but could now test the new function.

The code looks clean and the documentation sufficient.

Functionality-wise:
1. When running the the script with `snakemake -j1 create_empty_ts` and `zeros` in settings.yaml the files are created as expected with zeros in the series column.
When running the script again with `empty` the files are not created. Only deletion of existing files leads to new creation with NaNs in series column. The same applies the other way around.
Not sure if this is intended. 

2. The `var_name` column is filled with placeholders from the scenario data. I assume you indended this to provide an example parameter name ?

3. I'd recommend to write a small unit test.

MaGering:
Thanks for your review @chrwm!

> * When running the the script with `snakemake -j1 create_empty_ts` and `zeros` in settings.yaml the files are created as expected with zeros in the series column.
>   When running the script again with `empty` the files are not created. Only deletion of existing files leads to new creation with NaNs in series column. The same applies the other way around.
>   Not sure if this is intended.
> 

I introduced a variable `overwrite` in `settings.yaml` with commit https://github.com/rl-institut/oemof-B3/pull/325/commits/73160f56083d73e34e9627999a7c9be2e2722564. It let's you the option to overwrite the empty time series if set to true

> * The `var_name` column is filled with placeholders from the scenario data. I assume you indended this to provide example parameter names ?
> 

Yes, thanks for checking. This is intended. However the loop over scenarios has been broken, thank you @Stefanie08 for finding this! :) 
I fixed it with commit https://github.com/rl-institut/oemof-B3/pull/325/commits/dcdd9fc88953193d301a67aab83b6ffc0fec8d49. Now the respective time series is updated with each scenario and in the end all duplicates are dropped, so that only unique time series are saved. Also I switched to retrieve the start of the date time from the filter instead from the scenario specification with commit https://github.com/rl-institut/oemof-B3/pull/325/commits/74738b51ecc9107453ef0acce1276b9cc626c378.

> * I'd recommend to write a small unit test.

Yes, that would certainly be good. Unfortunately, we did not do that consistently in oemof-B3. Due to time constraints, I would not do that at this point.

MaGering:
@Stefanie08 could you test my changes? Try to see if anything else can be broken. :-D

With 
```
snakemake -j1 create_empty_ts -f
```
you can make a force run with `snakemake`. This means that output will be produced even if it is already there. Please check if it conforms to the logic behind overwrite in `settings.yaml`.

chrwm:
After pulling your latest changes, when running `snakemake -j1 create_empty_ts`, I get
```
The flag 'directory' used in rule create_empty_scalars is only valid for outputs, not inputs.
CRITICAL - The flag 'directory' used in rule create_empty_scalars is only valid for outputs, not inputs.
The flag 'directory' used in rule create_empty_ts is only valid for outputs, not inputs.
CRITICAL - The flag 'directory' used in rule create_empty_ts is only valid for outputs, not inputs.
Building DAG of jobs...
WARNING - Building DAG of jobs...
Nothing to be done (all requested files are present and up to date).
WARNING - Nothing to be done (all requested files are present and up to date).
Complete log: .snakemake\log\2023-04-24T181312.420508.snakemake.log
WARNING - Complete log: .snakemake\log\2023-04-24T181312.420508.snakemake.log
```

chrwm (approved):
Functionality works as expected.

Stefanie08:
Hello @MaGering,

After creating the empty ts files with  ```snakemake -j1 create_empty_ts -f``` and adding the new scenario keys as well as paths into the `scenario.yml` , the preprocessing of a single scenario is not working. The following error is being raised: 

```
MissingInputException` in rule build_datapackage in file ..\oemof-B3_issue_fixes\oemof-B3\snakemake_rules/build_datapackage.smk, line 14:
Missing input files for rule build_datapackage:
    output: results/2050-80-el_eff/preprocessed
    wildcards: scenario=2050-80-el_eff
    affected files:
        raw/time_series/empty_ts_load
        raw/time_series/empty_ts_efficiencies
        raw/time_series/empty_ts_feedin
```
Therefore the logic behind `overwrite` could not be tested. 
Am i missing a parameter?

PR #333, opened 2023-07-06
Fix/unstacking warning cols

### LINKED ISSUE #332: Warning that source and comment get lost after unstacking should only appear if they are not empty
The [following warning](https://github.com/rl-institut/oemof-B3/blob/63d01ad423ae1e77b28a42646206a64df0ed4653/oemof_b3/tools/data_processing.py#L1051-L1058) should only appear if there are values in the column `source` or `comment`. Otherwise you will be spammed with a warning unnecessarily.

MaGering:
Resolves #332

With this PR the logging warning `Caution any remarks in column '{col}' are lost after unstacking.` is only printed if column `source` or `comment` is not empty.

Open ToDos:

- [ ] Adjust the tests to check if the warning appears if the source or comment column is not empty and omits if they are empty.

### DIFF
--- a/oemof_b3/tools/data_processing.py
+++ b/oemof_b3/tools/data_processing.py
@@ -1053,7 +1053,7 @@ def unstack_timeseries(df):
     lost_columns = ["source", "comment"]
     for col in lost_columns:
         if col in list(df.columns):
-            if not _df[col].isna().any() or _df[col].values.any() == "None":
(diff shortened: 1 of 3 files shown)


MaGering:
Thanks for your work @Stefanie08 in implementing the tests!
Could you do a final review of this PR please? :) Thnx!

Stefanie08 (approved):
Thank you for the revision. Works as expected. :)

PR #335, opened 2023-07-27
Features/update tabular

### LINKED ISSUE #316: Python versions >3.8 not supported in github checks
I tried to update the python versions in `oemof-B3` to 3.8 and 3.9.

However the check with 3.9 failed because of [this error message](https://github.com/rl-institut/oemof-B3/actions/runs/4186026998/jobs/7253943566).

See PR https://github.com/rl-institut/oemof-B3/pull/313 for more info.

Stefanie08:
Resolves #316 

With this PR oemof-tabular is updated.


MaGering:
@Stefanie08 I used `concat` instead of `_append` in commit https://github.com/rl-institut/oemof-B3/pull/335/commits/34da9b08d67d4d1b5e101886ed55623aa236551b, since underscore methods are unofficial and might not be stable.

pyproject.toml
```diff
@@ -19,20 +19,21 @@ requires = ["poetry_core>=1.0.0"]
...
-oemoflex = { git = "https://git@github.com/rl-institut/oemoflex", tag = "0.0.1"}
+oemoflex = { git = "https://git@github.com/rl-institut/oemoflex", tag = "features/update_tabular"}
 Sphinx = {version = "^4.3.2", optional = true}
 sphinx-rtd-theme = {version = "^1.0.0", optional = true}
 sphinxcontrib-bibtex = {version = "^2.4.1", optional = true}
 docutils = "^0.17"
 geopandas = {version = "^0.10.2", optional = true}
 demandlib = "^0.1.8"
 pandas = ">=1.1.0"
+boto3 = "1.26.125"
```
MaGering, line 36: **Optional:** Check if this can be dropped:

1. Drop
2. Execute `poetry lock`
3. Revert in case poetry lock takes a disproportionately long time or other difficulties
4. In case poetry lock file is build after a short while execute poetry install and test all important oemof-B3 functionalities

If this can't be done due to time constraints, move this to a issue with milestone v0.0.4.

MaGering (changes requested):
Only the `oemoflex` version needs to be updated after `oemoflex` is released.
Besides that this PR is good to go!

pyproject.toml
```diff
@@ -19,20 +19,21 @@ requires = ["poetry_core>=1.0.0"]
...
 [tool.poetry.dependencies]
-python = ">=3.7, <3.9"
+python = ">=3.8, <3.11"
 numpy = ">=1.21.0"
 rtree = ">=0.9.7"
 pyyaml = ">=6.0"
 dynaconf = "^3.1.7"
 snakemake = ">=5"
-oemoflex = { git = "https://git@github.com/rl-institut/oemoflex", tag = "0.0.1"}
+oemoflex = { git = "https://git@github.com/rl-institut/oemoflex", tag = "dev"}
```
MaGering, line 28:
```suggestion
oemoflex = { git = "https://git@github.com/rl-institut/oemoflex", tag = "0.0.2a"}
```

pyproject.toml
MaGering: I dropped `boto3 = "1.26.125"`. We had it because the log file would not been created without it. I tested it and the workflow runs normally without it.

MaGering (approved):
Changes work as expected.

PR #336, opened 2023-08-29
Update labels

MaGering:
Resolves #328
Addresses #303

With this PR labels are fixed.

It does not close issue #303 since there are further errors running `snakemake -j1 plot_all_resources`.

PR #337, opened 2023-08-30
Replace deprecated op parameter with predicate

MaGering:
With this PR the Future warning 

```
sys:1: FutureWarning: The `op` parameter is deprecated and will be removed in a future release. Please use the `predicate` parameter instead.
```
is removed by using the parameter `predicate` instead of `op`.


MaGering:
Checks not working due to problem described in my comment [here](https://github.com/rl-institut/oemof-B3/pull/336#issuecomment-1697162784). Hence this PR should be merged after https://github.com/rl-institut/oemof-B3/pull/336.

MaGering:
Ready for review @Stefanie08!

Stefanie08 (approved):
Thank you for fixing the FutureWarning. The changes work as expected, although new FutureWarnings are raised. A new issue will follow.

MaGering:
> Thank you for fixing the FutureWarning. The changes work as expected, although new FutureWarnings are raised. A new issue will follow.

Thanks for checking @Stefanie08.

If I execute the rule which belongs to this PR

```
snakemake -j1 prepare_conv_pp
```
I do not get any further Future Warnings. Note that to test this rule, all files except `oemof-B3/results/_resources/scal_costs_efficiencies.csv` must be there.

If there are other `FutureWarning`s, you need to identify at which rule they occur. You can do this from the output in the console or the log file.
Please only create an issue in the `oemof-B3` repo if it is a `FutureWarning` that is triggered by code used in this repo. If it is a `FutureWarning` triggered in `demandlib` for example, you would have to create an issue there. Thanks a lot!

PR #340, opened 2023-08-30
Give UserWarning that report is not possible for windows user

### LINKED ISSUE #339: Report not working for windows users
The rule report is working with linux script. Windows not supported. This needs to be written down somewhere in the documentation or displayed for the user.

MaGering:
Resolves #339

With this PR the report is build for Linux or Darwin system users.
For windows users a `UserWarning` is printed that the report can't be build for windows users yet.
For all other systems (e.g. Java) a `UserWarning` is printed that the report can't be build for their system.


Stefanie08 (approved):
The expected `UserWarning` is raised for all scenarios (both dev and update_tabular env).

PR #357, opened 2023-11-23
Features/add tests

MaGering:
With this PR tests of the pipeline are added.

The aim is to test each of oemof-B3's rules. So far we collected the rules we had to test in this [excel table](https://wolke.rl-institut.de/f/521557). Now we want implement tests for this.

I already created a first draft, where I started by testing the rule "create_empty_scalars".
The basic idea is to check if the output, which would be created by the rule, already exists. If so it is renamed first. Then the rule is run and it is checked if there is an output from snakemake and if the path exists. If so, the created output is deleted and the original file is named back. There is a try and exception around this, to make sure, that original data is named back no matter what error there might occure from the renaming on.


MaGering:
@Stefanie08, here are some open TODOs:

- [x] Add docstrings to scripts
- [x] Add docstrings to functions
- [x] Add comments where needed

tests/test_pipeline.py
```diff
@@ -4,6 +4,19 @@
 
 
 def install_with_extra(extra):
+    """
+    This function installs extra packages stored in tool.poetry.extras
+
+    Inputs
+    -------
+    extra : str
+        Name of extra packages in a list.
```
MaGering: The info with the list is a bit confusing, since extra is a string. Maybe change to:
`Name of the list with the extra packages to be installed from pyproject.toml as string`?

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
 
     Outputs
     -------
-    absolute_path_list : str
-        Absolute file path
+    absolute_path_list : list
+         Absolute file path in list
```
MaGering: Here is one space too many.

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
 
+def file_name_extension(raw_file_path):
+    """
+    This function rearranges the current absolute file path
+    with the new extension suffix '_original'.
+
+    Inputs
+    -------
+    raw_file_path : str
+        Absolute path of rule
```
MaGering: This function could also be used regardless of whether it is raw data or a rule. Therefore, simply delete ‘of rule’ and rename ‘raw_file_path’ to ‘file_path’ in this function.

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+
+    Outputs
+    -------
+    renamed_path : str
```
MaGering: Could you rename the output to renamed_file_path for consistency reasons please?

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+    renamed_file = rename_path(
+        raw_file_path, "." + file_extension, "_original." + file_extension
+    )
+
+    return renamed_file
+
+
+def rule_test(sublist):
+    """
+    This function runs the rule from the output rule sublist.
```
MaGering: This function runs a specific rule with snakemake and asserts the runability with True

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+
+    Inputs
+    -------
+    sublist : str
+        Path of rule
```
MaGering: Change to "rule_path"

docs/conf.py
```diff
@@ -12,43 +12,44 @@
...
+sys.path.insert(0, os.path.abspath("../oemof_b3/config/"))
 
 # -- Project information -----------------------------------------------------
 
-project = 'oemof-B3'
-copyright = '2020, Reiner Lemoine Institut'
-author = 'Reiner Lemoine Institut'
+project = "oemof-B3"
+copyright = "2020, Reiner Lemoine Institut"
+author = "Reiner Lemoine Institut"
```
MaGering, line 24: Do you know why these and following changed?

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+
+
+def clean_file(sublist, delete_switch, renamed_path):
+    """
+    This function removes test data files and reverts renamed files.
+
+    Inputs
+    -------
+    sublist : list of str
+        List of target paths of rules
```
MaGering: Please rename "sublist" to "file_path_list"

List with absolute file path

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+    delete_switch : bool
+        If True, delete the data created during the test run.
+        If False, do not delete the data.
+    renamed_path : list of str
+        List of renamed target paths
```
MaGering: For consistency reasons:

    renamed_file_path_list : list of str
        List with renamed absolute file path

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+
+    Outputs
+    -------
+     None
+
+    """
+    # Remove the file created for this test
+    for raw_file_path in sublist:
```
MaGering: for file_path in file_path_list:

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
+                renamed_file,
+                "_original." + file_extension,
+                "." + file_extension,
+            )
+
+
 def pipeline_file_output_test(delete_switch, output_rule_list):
+    """
+    This function tests the Snakemake pipeline for a list of output rules
+    and reverts all changes made in the target directory.
```
MaGering: So far only for files and not for dirs. But when we are going to reuse it for dir tests as well in test_pipeline_results, we should rename the function instead.

oemof_b3/tools/testing_pipeline.py
```diff
@@ -163,83 +181,169 @@ def get_abs_path_list(output_rule_list):
...
-                # Get file extension
-                file_extension = raw_file_path[raw_file_path.rfind(".") + 1 :]
-                # Rename existing user data
-                renamed_file = rename_path(
-                    raw_file_path, "." + file_extension, "_original." + file_extension
-                )
-                renamed_path.append(renamed_file)
+            except FileNotFoundError as e:
+                print(e)
+                continue
```
MaGering, line 411: Where does this try and except come from?

tests/test_pipeline_raw_data.py
```diff
@@ -1,3 +1,7 @@
+"""
+This script checks the snakemake pipeline for target rules that create
+empty time series and scalars in the folder raw.
```
MaGering: And also the download from zenodo...

MaGering (commented):
Hey @Stefanie08,

could you please check my comments? Thank you! :)

oemof_b3/tools/testing_pipeline.py
```diff
@@ -0,0 +1,358 @@
...
+def remove_raw_data_created(exists):
+    raw_dir_path = get_raw_path()
+
+    if not exists:
+        if os.path.isdir(raw_dir_path):
+            shutil.rmtree(raw_dir_path)
+        else:
+            raise FileNotFoundError(
+                f"Something went wrong. {raw_dir_path} has been created but could not be found."
+            )
```
MaGering: Could you please add docstrings to these @Stefanie08? Thank you!! :)

tests/test_pipeline.py
Stefanie08: Changes have been adapted. Thank you for the fast review!

oemof_b3/tools/testing_pipeline.py
Stefanie08: I additionally changed the labeling for `_original` to extension as the `.csv` is already labeled as suffix.

oemof_b3/tools/testing_pipeline.py
Stefanie08: Thank you @MaGering for pointing this out! I will change the description accordingly.

I left the function `pipeline_file_output_test` as is, therefore the tests for the files stays unaffected. In the function `pipeline_folder_output_test` I adjusted the code to process both dirs and files. This needs to be tested for `test_pipeline_resources_tables.py` and `test_pipeline_raw_data.py` .

oemof_b3/tools/testing_pipeline.py
Stefanie08: I integrated a try block to process both dirs and files. The idea was to check if the path exists as file or dir and rename them. If the path does not exist but the path + _original, then it should return an error message. There is also the instance where the file does not exist at all and therefore should continue to the `rule_test`. I reverted the changes and put them into the `pipeline_folder_output_test` .

oemof_b3/tools/testing_pipeline.py
Stefanie08: The docstrings were added in the commit [4f93e74](https://github.com/rl-institut/oemof-B3/pull/357/commits/4f93e746157496b8da292f0ab1b828d56a611971).

oemof_b3/tools/testing_pipeline.py
Stefanie08: The function has been adjusted and renamed with this commit [ca75bbd](https://github.com/rl-institut/oemof-B3/pull/357/commits/ca75bbdc446f1cae4c97cdd0d985c2bbb8ee9cc5).
The function has been tested for `test_pipeline_resources_tables.py`, `test_pipeline_raw_data.py`, and `test_pipeline_results.py`.

oemof_b3/tools/testing_pipeline.py
Stefanie08: The function has been deleted with the commit [5e3eb48](https://github.com/rl-institut/oemof-B3/pull/357/commits/5e3eb48a6bddc114c941645784807940f40b4105).

oemof_b3/tools/testing_pipeline.py
```diff
@@ -0,0 +1,438 @@
...
+    output = snakemake.snakemake(
+        targets=rule_path,
+        snakefile="Snakefile",
+    )
+
+    # Check if snakemake rule exited without error (true)
+    assert output
+
+    # Log the success
+    logging.info(f"Snakemake rule executed successfully for targets: {rule_path}")
```
MaGering, line 286: @Alaadin17 : Triggers the rule test

tests/test_pipeline_raw_data.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+import snakemake
+import shutil
+from oemof_b3.tools.testing_pipeline import get_repo_path, pipeline_output_test
+from oemof_b3.config import config
+
+logger = config.add_snake_logger("data_processing")
+
+# Delete data from test run of pipeline if True otherwise False
+delete_switch = True
```
MaGering, line 14: @Alaadin17 please keep the delete_switch in your branch for the case developer wants to keep all test results.

MaGering:
Surpassed by https://github.com/rl-institut/oemof-B3/pull/365.

PR #365, opened 2025-10-14
Features/add tests reworked

MaGering:
This PR proposes a simpler version of the closed PR https://github.com/rl-institut/oemof-B3/pull/357

With this PR tests of the pipeline are added.

The aim is to test each of oemof-B3's rules. So far we collected the rules we had to test in this [excel table](https://wolke.rl-institut.de/f/521557). Now we want implement tests for this.

I already created a first draft, where I started by testing the rule "create_empty_scalars".
The basic idea is to check if the output, which would be created by the rule, already exists. If so it is renamed first. Then the rule is run and it is checked if there is an output from snakemake and if the path exists. If so, the created output is deleted and the original file is named back. There is a try and exception around this, to make sure, that original data is named back no matter what error there might occure from the renaming on.


oemof_b3/tools/testing_pipeline.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    # Raw data is needed for some rules and therefore is created if missing
+    raw_data_exists = check_raw_data_exists()
+
+    # TODO @ Alaadin17: Implement here: If results_path (= dir "results") not empty, raise:
+    #         raise FileExistsError(
+    #             f"The directory {results_path} is not empty. \n"
+    #             f"The test can not be executed. Please delete all files in {results_path} first
+    #             and then execute again."
+    #         )
```
MaGering: @Alaadin17: Could you please take over this task?

tests/test_pipeline_results.py
```diff
@@ -0,0 +1,82 @@
...
+    _scenario_output_rule_list = [
+        "results/" + scenario + "/preprocessed",
+        "results/" + scenario + "/optimized",
+        "results/" + scenario + "/postprocessed",
+        "results/" + scenario + "/b3_results/data",
+        "results/" + scenario + "/tables",
+        "results/" + scenario + "/plotted/dispatch",
+        "results/" + scenario + "/plotted/storage_level",
+        "results/" + scenario + "/plotted/scalars",
+        # TODO @ Alaadin17: Add rule check of esys graph
```
MaGering: @Alaadin17: Could you please integrate the test of the esys graph?

oemof_b3/tools/testing_pipeline.py
MaGering: Resolved with https://github.com/rl-institut/oemof-B3/pull/365/changes/e0d8d2557ec144ac73eb8ff6d08db4aa22a1f965

tests/test_pipeline_results.py
MaGering: Resolved with https://github.com/rl-institut/oemof-B3/pull/365/changes/f04f6be1508874a4e3888d0d266351185c241258

